PR #3, opened 2020-01-22
[RFC] Richardson error mitigation

eggerdj:


eggerdj:
Migrated from https://github.com/Qiskit/qiskit/pull/768

text/####-error-mitigation.md
```diff
@@ -0,0 +1,235 @@
...
+- It does not include single-qubit gates.
+- The effective stretch factors to chose from are very limited.
+- Some two-qubit gates, such as root-SWAP, need to be applied more than twice to compose to the identity.
+
+### Backend constrained error mitigation
+When implementing error mitigation using stretch factors, as is done in https://arxiv.org/abs/1805.04492, new pulses must be defined and calibrated for the different stretch factors c_i.
+To implement error mitigation in the manner the backend could have a set of pre-defined calibrated pulses with different stretch factors.
+For instance, following Kandala et al., the backend could store calibrated pulses for c=1 (i.e. the pulses used in regular operations), c=1.1, c=1.25, and c=1.5.
+At execute time, the user would specify that he wants to run a quantum circuit using error mitigation.
+The pulse scheduler would then create four copies of the quantum circuit, each with a different stretch factor supported by the backend.
```
ajavadia: Further down you say this is the responsibility of the assembler. I think that's correct. The pulse scheduler should not be involved in any of this.
```suggestion
The assembler would then create four copies of the quantum circuit, each with a different stretch factor supported by the backend.
```

text/####-error-mitigation.md
```diff
@@ -0,0 +1,235 @@
...
+First, we define that a given qubit has stretch factor `c` available if all gates that involve this qubit are available with this stretch factor.
+The user will thus be able to query the backend to find out which stretch factors are available on a given list of qubits:
+```
+def available_stretch_factors(self, qubits: List) -> List:
+    """Returns a list of stretch factors available for the given qubits."""
+```
+For example to find the stretch factors available to qubits 0, 1, and 3, the user would do `backend.available_stretch_factors([0, 1, 3])`.
+Similarly, the user may query the backend to find out which qubits support a given list of stretch factors.
+```
+def available_qubits(self, stretch_factors: List) -> List:
```
ajavadia: To use this function one should know apriori what stretch factors there are. So I'm not sure it provides a lot of value over the methods the previous method.

text/####-error-mitigation.md
```diff
@@ -0,0 +1,235 @@
...
+
+print(qobj.experiments[4])
+{
+    'config': {...},
+    'header': {'name': 'circB_ec01b', 'stretch_factor': '1.1', ...},
+    'instructions': {...}
+}
+```
+Here, the instructions for experiments 0, 1, and 2 are identical since they correspond to `circA`.
+The instructions for experiments 3, 4, and 5, are also identical since they correspond to `circB`.
```
ajavadia: Seems like the above code snippet should include 5 experiment examples.

text/####-error-mitigation.md
```diff
@@ -0,0 +1,235 @@
...
+```
+This implies that the header of the `QasmQobjExperiment` will need to be modified to include an attribute, namely the `stretch_factor`.
+
+### Backend execution
+The backend will have to be able to detect that stretch factors are present in the header of the `qobj.experiments`.
+This will then signal to the backend what type of gates to use to execute each circuit.
+
+### Result returned by the backend
+The result returned by the backend would contain the results for the circuits executed with different stretch factors.
+This result object could then be given to a Richardson error mitigation module in Qiskit Ignis that would perform the extrapolation to the zero-noise limit.
```
ajavadia: Ignis would have to understand from the Result object which ensemble of `ExperimentResult` belong to one error mitigation. Is the idea that it would do this based on those headers/experiment names? It would be helpful to add a bit of detail on that as well, maybe following the example of the Qobj above.

text/####-error-mitigation.md
eggerdj: Yes you are correct. Thanks.

text/####-error-mitigation.md
eggerdj: You are correct. I think `available_stretch_factors` will be more useful. I suggested adding `available_qubits` for completeness but we can also decide not to implement it and only implement it if we see a need for it later on.

text/####-error-mitigation.md
eggerdj: Why 5? Two circuits and three stretch factors results in 6 experiments. Of the six experiments, four are displayed. I'll add a bit more to the text to explain this more.

text/####-error-mitigation.md
```diff
@@ -0,0 +1,235 @@
...
+execute(circ, backend, ..., error_mitigation='richardson', stretch_factors=[1.0, 1.25, 1.5]).
+```
+This will require a validation step, done in Qiskit Terra and based on the `config`, to check that the specified stretch factors are available for all of the qubits in the quantum register of `circ`.
+If this is not the case, an error will be raised.
+Here, `error_mitigation` specifies the error mitigation method to use.
+This value would by default be `None` when error mitigation is not used.
+The changes needed in Qiskit to implement error mitigation would require changes in the `assemble` function.
+`assemble` would detect that the user requires error mitigation and replace each circuit by several circuits, one for each stretch factor.
+The `experiments` in the resulting `qobj` would have an entry in the header that specifies the stretch factor.
+For example, the `qobj` for two circuits (`circA` and `circB`) to be run with error mitigation with three stretch factors would look like
```
eggerdj:
```suggestion
For example, the `qobj` for two circuits (`circA` and `circB`) to be run with error mitigation with three stretch factors would include six experiments, four of which would look like
```

eggerdj: @ajavadia Is this what you were getting at in the question below?

text/####-error-mitigation.md
eggerdj: Good idea.

eggerdj:
```suggestion
This result object could then be given to a Richardson error mitigation module in Qiskit Ignis that would perform the extrapolation to the zero-noise limit.
This module would need to know which result is associated to which quantum circuit and stretch factor.
This may be done by adding the required stretch factor as a parameter in the header of the result. For the example above, with `circA` and `circB`, the backend may return the `mitigated_result` in which the header of each result may look like

print(mitigated_result.results[0].header)
Obj(name='circA', stretch_factor='1.0')
print(mitigated_result.results[1].header)
Obj(name='circA', stretch_factor='1.1')
print(mitigated_result.results[3].header)
Obj(name='circB', stretch_factor='1.0')
print(mitigated_result.results[4].header)
Obj(name='circB', stretch_factor='1.1')

```

text/####-error-mitigation.md
ajavadia: yes

text/####-error-mitigation.md
ajavadia: sounds good

text/####-error-mitigation.md
ajavadia: yes 6 is correct

text/####-error-mitigation.md
```diff
@@ -0,0 +1,247 @@
...
+```
+To facilitate the user experience, helper functions that show the available stretch factors, will be created.
+First, we define that a given qubit has stretch factor `c` available if all gates that involve this qubit are available with this stretch factor.
+The user will thus be able to query the backend to find out which stretch factors are available on a given list of qubits:
+```
+def available_stretch_factors(self, qubits: List) -> List:
+    """Returns a list of stretch factors available for the given qubits."""
+```
+For example to find the stretch factors available to qubits 0, 1, and 3, the user would do `backend.available_stretch_factors([0, 1, 3])`.
+Similarly, the user may query the backend to find out which qubits support a given list of stretch factors.
```
eggerdj:
```suggestion
Similarly, the user may query the backend to find out which qubits support a given list of stretch factors.
The following function would implement this functionality
```

text/####-error-mitigation.md
```diff
@@ -0,0 +1,247 @@
...
+```
+def available_qubits(self, stretch_factors: List) -> List:
+    """Returns a list of qubits for which the given stretch factors are available."""
+```
+The user may also query the backend to find out how the error mitigation is configured.
```
eggerdj:
```suggestion
However, the user would need to know a priori which stretch factors are available on the backend.
We would, therefore, only implement this function later on if we see a need for it.
The user may also query the backend to find out how the error mitigation is configured.
```

PR #10, opened 2020-01-28
[RFC] Aqua's QuantumAlgorithm: Interfaces, Instance Reuse, and Organization

Cryoris:
This WIP RfC discusses the contents and necessity of Aqua's `QuantumAlgorithm`, which is the base class for all quantum algorithms.

### DIFF
--- a/####-aqua-quantum-algorithm.md
+++ b/####-aqua-quantum-algorithm.md
@@ -11,10 +11,7 @@
 
 ## Summary
 
-Aqua's quantum algorithms are of type `QuantumAlgorithm`, which dictates that all algorithms must be executed via a  `run` method whose arguments specify the backend used for circuit execution. This structure makes passing algorithm instances to meta algorithms difficult and hacky as once instantiated the problem the algorithm solves is essentially fixed.
-
-This RFC discusses the issues of the current `QuantumAlgorithm` and tries to assess, whether
-this base type is still useful or if it might complicate the development of quantum algorithms.
+Aqua's quantum algorithms are of type `QuantumAlgorithm`, which dictates that all algorithms must be executed via a  `run` method whose arguments specify the backend used for circuit execution. This structure makes passing algorithm instances to meta-algorithms difficult and hacky as once instantiated the problem the algorithm solves is essentially fixed. This RFC discusses the issues of the current `QuantumAlgorithm` and tries to assess, whether this base type is still useful or if it might complicate the development of quantum algorithms.
 
 The below proposal introduces the following:
 1. **Interface Classes** - Non-mutually-exclusive base classes for various types of algorithms, such as MinEigenSolver, ClassicalAlgorithm, Classifier, or AmplitudeEstimator, enforcing a consistent execution signature (`get_min_eigenvalue()`) which is still flexible enough to support diverse execution flows for different algorithm _types._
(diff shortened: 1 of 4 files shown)


woodsp-ibm:
Aqua has been in need of some better type definition in regards to behavior and result across algorithms to allow them to be better used interchangeably. While input types (no longer present as these were more for declarative) the output started as dictionaries but the intent was to have result objects/types at some point in time. The problem/behavior was previously by name in the configuration so say an algorithm that could handle an 'Ising' problem was expected to provide certain dictionary fields as output though it could supply more if there were algorithm specifics of interest. And as such only related to output data and not any methods on algorithms. In the model as it stood and algorithm was created and run and the result was used by the caller.

For an algorithm to be substituted the minuimum is the input data would be expected to be the same, as well as the output result. Configuration can be different assuming a knowing third party constructs the algorithm as needed and then supplies it for use in a manner that the receiver expects and can handle - e.g. finding min eigenvalue, finding excited state(s).

Having an algorithm implement one or more different problem 'classes' proving said standard behavior seems reasonable. How much of this might belong to the class versus directly as methods and what might be part of a result object depends on design and function. Clearly an algorithm should be able to used for different problems and implementing multiple classes must be possible with no conflict. As to these "MinEigenSolver, ClassicalAlgorithm, Classifier, or AmplitudeEstimator" I think as an example ClassicalAlgorithm is more a how it does it and is not useful in this context, the other 3 seem to be much more what its doing and seem more valid as 'interface' classes to express the problem/behavior.

The QuantumAlgorithm was designed as base class for common function and behavior. It used to have a lot of circuit related logic but this was all moved out. It was done both as a convenience for a developer of a new algorithm as well as providing a common functionality to external users which is pretty much the run() method. If algorithms are going to be broken into sub-types, and used on the basis of the sub-type, which is pretty much what was happening in the past, then maybe there is less need for an overall base class. As to train, test and predict, yes the classifiers, like SVM can be limited to that - i.e. remove the run, when we have 'interface' types that express behavior.

Incomplete construction. This needs to be spelled out to end user what this is. I don't think we can expect any help from porgramming language here. E.g an algorithm taking an operator I would assume now it would say operator=None in contructor and then in docstring state that if not supplied MUST be supplied later on solve(operator=None) method which will override any operator supplied earlier. If its never supplied I guess the algorithm should check for that now, and raise a meaningful exception that effect, as neither method can require it like before. 

Reuse puts extra burden on algorithm developers to ensure that any instance variables that say are used for state and control of algorithm behavior when 'running' are reset on run/solve or whatever its called before the algorithm is executed. Setting them up in constructor will NOT be enough for reuse in this case. If we support re-use then tests must be in place to ensure that it works. The alternative is of course for the initial instance to be cloned externally by the user and the clone used. Doing deepcopy may be expensive (or not) depending on what is instantiated early inside the algorithm.


> Algorithms are fixed, the problems they solve are not. Why do we have to specify a new algorithm instance for each problem when we could just input the information when we actually run it?

It was expected that creating a new instance was lightweight compared to other processes. If one does not know how to create then of course it could be cloned. Re-use means more careful development as the constructor may now not the only place that initialization needs to take place - if you want to re-run and it has internal state then that state needs resetting. Aguably more to go wrong.


> Missing flexibility: configuring the algorithm in the initializer makes the algorithm instance problem-specific. We are therefore unable to pass an algorithm instance that can solve different problems to meta algorithms. Examples: ADMM, any optimization, etc.

I think the flexibility you are talking here relates simply to common data input for the problem. Overall configuration amy be very different unless you can define some common methods on 'interface' classes that can convert from some generic to teh specifics of the algo.

> The term "algorithm" is extremely broad. Classical algorithms don't share a common base class (compare Qiskit, scikit, etc.), why should quantum algorithms share one?

They did so for common function as explained above.

> Naming: every single quantum algorithm has is executed using `run`, though other names can be much more expressive. Optimizers `optimize` or `minimize` a quantity, neural networks `learn`, HHL `solve`s, etc. Classical algorithms do not have this naming restriction, why should quantum algorithms have them?

It was done for consistency as explained above. If we have 'interface' classes that can express function maybe it can be done via these. Though algorithms then may implement multiple methods.

> If an algorithm runs and compares multiple backends or settings, this is not supported in a convenient manner.

An algorithm is run on a backend under settings however it is configued. To compare different runs this needs to done externally to maintain mutiple run outcomes. I do not see this as part of an algorithm.

> Algorithms can be hardware specific, the backend must be known beforehand and it doesn't make sense to provide the backend in `run`.

Any instantiation of hardware dependent logic can be and was delayed until the backend was known. It's not really needed until its run anyway - unless you want to see hardware specific ccts ahead of running and then we did allow a user to supply that. It was a requirement from the top to make algorithms largely independent of backend in the same way that QuantumCircuit is independent of backend at the time of creation.


I am not going to comment on the 'benefits' here as described in the RFC aside from this. This makes things more flexible and varied. It better facilitates use of different algorithms in a common way. Arguably it makes more things more complicated as there are more types, methods etc, partials and re-use will require care by developers etc. 

As to directory heirarchy we had always favored flat. No matter how you categorize them, and people want to do it differently, they can only be in one place - documentation on other hand can list them under multiple categories. It seemed there were relatively few algorithms so flat was not exactly an issue. I an fine with having classical separated out it seems best that way - but otherwise origanized based on problem solved, since some algorithms solve multiple problems that seems less obvious to me. (I'm thinking here around min eig val, excited states, ising that we have various algorithms that handle one or more of these) If folders/categories are thought to work then the proposal should show all algorithms - new ones we would have to deal with as they arise of course.

Result storage. I believe each algo today builds up the dictionary internally and returns it all. It is still there in the algo, assuming its not re-run. Whatever the result object is I think must suffice. It can contain methods to compute values based on data in the result. Nothing short of re-running the algo should cause other/more results to be computed other than what can be computed based on the result data itself. This seems especially so when it must excute ccts etc and potentially end up with slighly different outcome. Just store the entire result and hence if you forget the you have the full result. If you want to have the QuantunAlgorithm hold the last 'n' results for history, that could be done. Lets not make things too complicated!

It seems that this is moving backend/QuantumIntance type logic around. Maybe its not too bad as such and any change will not impact overly whatever might arise over the execution environment change thought.

---------------------

Lastly. As an in interim, if some of this is needed in short term. Might a short term alternative be to put in place some wrapper for the classes we have today that provides this common function. Much in the same way scipy optimize is a wrapper to the underlying optimizers and presents a common interface. Yes this might be limited to the current set of algos and for the longer term changes to the current code may be better than wrappers. But it may be more manageble short term depending on exact short term needs.

PR #11, opened 2020-01-30
[RFC] Aqua 0.7: Quantum Optimization Stack

Cryoris:
This RFC discusses the quantum optimization stack, planned for the Aqua 0.7 release.

### DIFF
--- a/####-optimization-stack.ipynb
+++ b/####-optimization-stack.ipynb
@@ -1,0 +1,333 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Quantum Optimization Design Doc"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Motivation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Current Situation\n",
+    "\n",
+    "#### Current restriction to QUBOs\n",
+    "Currently, Qiskit only supports Quadratic Unconstrained Binary Optimization (QUBO) problems.\n",
+    "QUBOs can be mapped to an equivalent operator ground state problem.\n",
+    "The corresponding algorithms available in Qiskit (VQE / QAOA) expect an 'Operator' in the constructor that specifies the problem to be solved.\n",
+    "\n",
+    "For some problem classes (e.g. TSP, MaxCut, Portfolio, ...), Qiskit provides dedicated translators that take problem parameters and return the corresponding operator.\n",
+    "In addition, a more general translator from a `docplex.mp.model.Model` to an `Operator` is available.\n",
+    "\n",
+    "In principle, `docplex.mp.model.Model` can represent more general types of (quadratic) optimization problems, but since the available algorithms require an `Operator` as input, we are again limited to QUBOs.\n",
+    "\n",
+    "#### Missing model flexibility for advanced algorithms\n",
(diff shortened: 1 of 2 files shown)


woodsp-ibm:
Some comments, most of it is around detail more than the big picture which is pretty much as expected.

* Class names OptimizationProblem, OptimizationAlgorithm seem ok, maybe instead of Solution - OptimizationResult.

  "qiskit/optimization/solutions    # contains `Solution` and related quality metrics"

  (thinking has been for some time Aqua Algos will extend a generic AlgorithmResult for specific return
  types (results i.e. output) instead of "plain" dictionaries - note named result). 
 
* converters and translators. Currently when ising convertors were refactored 'translators' was dropped since they were now in translator layer a.k.a application stack. Either way we should consider naming and what goes where as converters and translators since naming seems perhaps too closely related for ease of distinction. Ising under finance should then be made consistent in naming as needed.
  
  converters: translating a QUBO given as OptimizationProblem to an operator
  translators: problem-specific translators (to an operator)
  
  The ising convertors - do we want to keep them (or at least one or two? as direct examples on how this is done). Do others become a set of models we supply for given problems

* So far 'algorithms' is a name that has been limited to quantum algorithms that ran under Aqua. There are algorithms in Chemistry but these are really Aqua level just specific to chemistry. Maybe there is no better naming.

* Solution
  "A Solution object should contain the following fields (but very likely a lot more, depending on the algorithm):
    x (= final solution)
    f_mean"

  Will the end user be able to define a problem abstractly and have the result for said problem provided?  Reason for asking - today we can define a max-cut problem for docplex and get back an operator. Use that operator and get back some algorithm result, but then to 'interpret' in the context of the defined problem notebooks use the pre-built max-cut translator to do this.

* Comments on code samples (not sure how much they are simply illustrative that the comments do not really apply as such)

  - op.variables.add(names=["x0", "x1", "x2", "x3"], types='B'*4)
    `types = 'B'`; Docstrings can of course list valid 'types' might 
	enums (although perhaps more verbose, allow better IDE discovery)

  - SparsePair(ind=[0, 1, 2], val=[1.0, -2.0, 0.5]
    Choose meaningful names of params.
	ind == indices? Seems too like a candidate for default value 
	
  - These are all what - wrappers to Aqua algorithmsm new Aqua algos, combination thereof? 
    rqaoa = RQAOA(...)
    gas   = GroverAdaptiveSearch(...)
    cplex = CPLEX(...)
	
  - Partial execution - how far does this extend. Asked for of Aqua algos in another RFC but seems it could be done at this level too  
   	
       So as an example instead of doing (as per rfc)

	\# construct different versions of ADMM
        admm_rqaoa = ADMM(convex_solver, rqaoa, ...)
        admm_gas   = ADMM(convex_solver, gas, ...
        admm_cplex = ADMM(convex_solver, cplex, ...)
	\# test different version on problem
	res_rqaoa = admm_rqaoa.solve(op)
	res_gas   = admm_gas.solve(op)
	res_cplex = admm_cplex.solve(op)
	
	it could be something  like this re-using ADMM algo
	
	\# construct ADMM
        admm = ADMM(convex_solver, ...)
	\# test different versions on problem
	res_rqaoa = admm.solve(op, rqaoa)
	res_gas   = admm.solve(op, gas)
	res_cplex = admm.solve(op, cplex)
	
	and/or alternatively (setter versus params on run/solve method)

	admm.set_convex_solver(rqaoa)
	res_rqaoa = admm.solve(op)
	admm.set_convex_solver(gas)
	res_gas = admm.solve(op)
	...

PR #13, opened 2020-02-05
[RFC] Encoding of integer variable with binary variables (Aqua optimization)

t-imamichi:
I propose a framework of converter of `OptimizationProblem` and encoding of integer variable with binary variables for Qiskit Aqua optimization stack.


t-imamichi:
This should be implemented as part of #11.

PR #8, opened 2020-01-24
[RFC] Aqua 0.7 Operator Refactor Design Doc

dongreenberg:
The Operator Redesign doc, copied over from https://github.com/Qiskit/qiskit-aqua/pull/742. I still need to tweak the doc to reflect the changes below, so please hold off on merging just yet.

Design Review Notes (Action Items marked by **AI**):

*   Initial Impressions
    *   Julien (@Cryoris )
        *   Good - Operators are designed to be more flexible, and easier to manipulate knowing the math
        *   Likes using the same syntax as in Terra, and easy access to the matrix
        *   Likes simplified call to expectation and evolution
        *   Questions:
            *   What happens if you combine combinations?
            *   Might prefer a flat hierarchy because of parallels to Composite gate in Terra
    *   Panos (@pbark )
        *   Good - Overall happy with doc, In favor of positioning closer to Terra
        *   Questions
            *   How to do rotated operator bases (e.g. Pauline's work)?
    *   Stefan (@stefan-woerner )
        *   Good: Structure is nice
        *   Syntactic sugar is too much for his taste, would rather keep everything explicit in the source code
        *   Agrees with Julien about combinations being tricky
    *   Yael (@yaelbh )
        *   Syntactic sugar may be too tricky
        *   Thinks primitives should be in Terra
        *   Important to find a solution to the AerExpectation magic branching in Aqua today
    *   Steve (@woodsp-ibm )
        *   Impression: Need to focus on Operator redesign, because the doc will take too long to implement, and we need to focus and be incremental.
        *   Still not clear - how do Expectation and Evolution algos play with the operators?
        *   Math between operators - unclear how conversion between types will be affected for combinations
    *   Bryce - Listener
    *   Paul - Listener
    *   Chris (@chriseclectic )
        *   Generally agree
        *   Needs to be a clear separation between primitive operators which are useful elsewhere and should be in Terra, and what’s just for Aqua
        *   Syntactic sugar should be optional, he doesn’t like it but some people do
        *   Separating between a data structure, what’s a numeric algorithm on the data structure, and what’s a synthesis algorithm on a composite structure
        *   Similarly for conversion between operators, some can’t and can’t be done, and some are exponential or expensive
        *   Agrees something needs to be done for Expectation or Evolution between algos and backends
    *   Kevin (@kdk )
        *   Agrees with Chris about delineation between primitives in Terra and high level in Aqua
        *   Joint ownership is the right model
    *   Ivano - Listener
*   Donny: Themes and questions to answer
[...]

PR #14, opened 2020-02-07
[WIP][RFC] Grouping of Pauli strings with entangled measurements (Aqua)

t-imamichi:
This RFC proposes to extends `qiskit.aqua.operators.tpb_grouped_weighted_pauli_operator` to provide grouping of Pauli strings with not only tensor product basis but also entangled measurements.


t-imamichi:
This should be implemented based on the operator class redesign #8.

PR #15, opened 2020-02-11
[RFC] Registerless circuits and circuit composition

ajavadia:
We propose simplifications to Qiskit circuits:
1. Registers are completely optional. They can be added where it makes sense. The qubits/clbits are the fundamental data, not registers.
2. The above makes it easier to compose circuits with each other, so you no longer need to convert circuits to instructions to anonymize them before appending onto another circuit.
3. All of the above makes it easier to treat gates/composite gates/subcircuits/circuits as the same thing.

### DIFF
--- a/circuits_over_bits.md
+++ b/circuits_over_bits.md
@@ -102,13 +102,14 @@ circuit = QuantumCircuit(6)   # a register-less circuit over 6 qubits
 print(circuit.qubits)
 >> [Qubit(), Qubit(), Qubit(), Qubit(), Qubit(), Qubit()]
 print(circuit.qregs)
->> None
+>> {}
 
 q = circuit.qubits
 circuit.mark_register([q[0], q[1], q[2]], 'a')  # now tag the bits as an 'a' and 'b' register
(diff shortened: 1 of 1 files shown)


circuits_over_bits.md
```diff
@@ -0,0 +1,199 @@
...
+>> {}
+
+q = circuit.qubits
+circuit.mark_register([q[0], q[1], q[2]], 'a')  # now tag the bits as an 'a' and 'b' register
+circuit.mark_register([q[2], q[3], q[4]], 'b')  # (this info will be kept in circuit)
+print(circuit.qregs)
+>> {'a': QuantumRegister(3, 'a'),
+    'b': QuantumRegister(3, 'b')}
+
+circuit.mark_register([q[0]], 'c')  # marking a bit as belonging to more than one register is disallowed
```
maddy-tod: In the example above `q[2]` is marked as belonging to both registers `'a'` and `'b'`, was it meant to be `circuit.mark_register([q[3], q[4], q[5]], 'b')`?

ajavadia: thanks, fixed

PR #16, opened 2020-02-20
[RFC] Qiskit Pulse: Relative timing for Schedules

lcapelluto:
Add an RFC proposing an implementation for relatively-timed instructions in pulse Schedules within Qiskit Pulse

### DIFF
--- a/0000-qiskit-pulse-schedule-relative-timing.md
+++ b/0000-qiskit-pulse-schedule-relative-timing.md
@@ -53,11 +53,12 @@ Factors to consider:
 - Compatibility -->
 
 ### Notation:
-I am going to use this notation here:
+I am going to use this notation here, where `self` is the `Schedule` initiating the operation in question:
 
- - |C| : the total number of channels in `self` (of type `Schedule`)
+ - |C| : the total number of channels in `self`
  - |N<sub>self</sub>| : the total number of instructions in `self`
- - |N<sub>other</sub>| : the total number of instructions in another `Schedule` being added in some way to `self`
+ - |N<sub>new</sub>| : the total number of instructions in another `Schedule` being added in some way to `self`
+ - max(|N<sub>C</sub>|) : the maximum number of instructions on any single channel within `self`
 
 ### New Schedule methods
 
@@ -75,7 +76,7 @@ I am going to use this notation here:
 
 | Method        | Runtime        |  Change          |
 |---------------|----------------|------------------|
-| `append(self, schedule)` | O(|N<sub>new</sub>|) |  |
+| `append(self, schedule)` | O(\|N<sub>new</sub>\|) |  |
 |`insert(self, time, schedule)`|  |  |
 
 ## Detailed Design
--- a/0000-qiskit-pulse-schedule-relative-timing.md
+++ b/0000-qiskit-pulse-schedule-relative-timing.md
@@ -11,7 +11,7 @@
 ## Summary
 Update the internal structure of `Schedule`s to track instructions sequentially. `Instruction`s will not be played at an explicit time; rather, they will begin immediately after the last is done. To schedule idle time between instructions, `Delay` instructions must be used.
 
-The user API will be updated to take advantage of this relative timing, with several new methods: `barrier`, `left_align`, `center_align`, and `right_align` (see: [new methods](#new-schedule-methods)).
+The user API will be updated to take advantage of this relative timing, with several new methods: `barrier`, `left_align`, `center_align`, and `right_align`.
 
 ## Motivation
 (This section was copied from [this issue](https://github.com/Qiskit/qiskit-terra/issues/3749), written by @taalexander.)
--- a/0000-qiskit-pulse-schedule-relative-timing.md
+++ b/0000-qiskit-pulse-schedule-relative-timing.md
@@ -147,7 +147,6 @@ A rough sketch for the implementation of some of the methods described above cou
  
 ## Alternative Approaches
 
-\# TODO
 
 ## Questions
 
--- a/0000-qiskit-pulse-schedule-relative-timing.md
+++ b/0000-qiskit-pulse-schedule-relative-timing.md
@@ -77,7 +77,7 @@ I am going to use this notation here, where `self` is the `Schedule` initiating 
 | Method        | Runtime        |  Change          |
 |---------------|----------------|------------------|
 | `append(self, schedule)` | O(\|N<sub>new</sub>\|) |  |
-|`insert(self, time, schedule)`|  |  |
+|`insert(self, time, schedule)`| O(\|N<sub>new</sub>\| * \|N<sub>C</sub>\|) |  |
 
 ## Detailed Design
 
@@ -91,13 +91,9 @@ I am going to use this notation here, where `self` is the `Schedule` initiating 
 \# TODO
 
 ## Questions
- 1. Issue with not breaking `x += y << t`:
 
-    `shift` will be easy to implement, but relative timing makes this impossible: `sched += sched2 << time`!!
-
-    Should be fine for `x |= y << t`.
  1. Should we create `insert(self, index, channel)`?
- 1. We can support the current `insert` method, but the runtime will be the maximum number of instructions on one channel, per instruction added. This will be incredibly slow for `x |= y << t`.
+ 1. We can support the current `insert` method, but the runtime with be O(|N<sup>2</sup>|).
  1. Is it best to remove leading/trailing `Delay`s on `left_align`/`right_align`, respectively?
 
 ## Future Extensions
--- a/0000-qiskit-pulse-schedule-relative-timing.md
+++ b/0000-qiskit-pulse-schedule-relative-timing.md
@@ -56,20 +56,14 @@ Factors to consider:
 I am going to use this notation here:
 
  - |C| : the total number of channels in `self` (of type `Schedule`)
-
- |N<sub>self</sub>|
- |N<sub>other</sub>|
+ - |N<sub>self</sub>| : the total number of instructions in `self`
+ - |N<sub>other</sub>| : the total number of instructions in another `Schedule` being added in some way to `self`
 
 ### New Schedule methods
 
- - `barrier(self, channels)`: O(|C|)
-
-   Add `Delay`s across `channels` so that all of the `channels` are occupied until `self.duration(channels)`.
-
-Should the following take `channels`, or should we "keep it simple, stupid"? A user could always build a subschedule of the channels, or use filter to get the schedule they want....
-
 | Method        | Runtime        |  Description     |
 |---------------|----------------|------------------|
+|`barrier(self)`| O(\|C\|) | Add `Delay`s across `channels` so that all of the `channels` are occupied until `self.duration(channels)`. |
 |`left_align(self)` | O(\|C\|) | This is the default behavior. We could have it return `self`, or we could remove any pre-pended `Delay`s. The latter is probably better, because `schedule.right_align().left_align()` should yield the original `schedule`. Should we also append `Delay`s? Likely. *Needs more thought* |
 |`center_align(self)` | O(\|C\|) | Pre- and ap-pend equally sized `Delay`s (how to split odd durations?) on each channel so that all channels are occupied until `self.duration()` |
 |`right_align(self)` | O(\|C\|) | Prepend `Delay`s on each channel so that all channels are occupied until `self.duration()`. Should we also remove trailing `Delay`s? |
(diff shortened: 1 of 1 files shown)


0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -0,0 +1,105 @@
...
+ - |N<sub>new</sub>| : the total number of instructions in another `Schedule` being added in some way to `self`
+ - max(|N<sub>C</sub>|) : the maximum number of instructions on any single channel within `self`
+
+### New Schedule methods
+
+| Method        | Runtime        |  Description     |
+|---------------|----------------|------------------|
+|`barrier(self)`| O(\|C\|) | Add `Delay`s across `channels` so that all of the `channels` are occupied until `self.duration(channels)`. |
+|`left_align(self)` | O(\|C\|) | This is the default behavior. We could have it return `self`, or we could remove any pre-pended `Delay`s. The latter is probably better, because `schedule.right_align().left_align()` should yield the original `schedule`. Should we also append `Delay`s? Likely. *Needs more thought* |
+|`center_align(self)` | O(\|C\|) | Pre- and ap-pend equally sized `Delay`s (how to split odd durations?) on each channel so that all channels are occupied until `self.duration()` |
```
taalexander: Do we include preexisting delays when centering?

0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -0,0 +1,105 @@
...
+
+
+### API Changes
+
+\# TODO
+
+| Method        | Runtime        |  Change          |
+|---------------|----------------|------------------|
+| `append(self, schedule)` | O(\|N<sub>new</sub>\|) |  |
+|`insert(self, time, schedule)`| O(\|N<sub>new</sub>\| * \|N<sub>C</sub>\|) |  |
```
taalexander: How do you expect inserting in the middle of a schedule work? Verify that delays are present for all insertion locations?

0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -0,0 +1,105 @@
...
+ - How new and existing methods will be updated
+
+ 
+## Alternative Approaches
+
+\# TODO
+
+## Questions
+
+ 1. Should we create `insert(self, index, channel)`?
```
taalexander: Isn't this handled by the normal insert, I don't see what is different about this?

0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -0,0 +1,105 @@
...
+ 1. We can support the current `insert` method, but the runtime with be O(|N<sup>2</sup>|).
```
taalexander: I think this would be useful, but we need to be clear that this is expensive.

0000-qiskit-pulse-schedule-relative-timing.md
lcapelluto: the normal insert takes `time`, this takes `index`. I don't really see how this `index` method would be useful myself

0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -0,0 +1,105 @@
...
+ - |C| : the total number of channels in `self`
+ - |N<sub>self</sub>| : the total number of instructions in `self`
+ - |N<sub>new</sub>| : the total number of instructions in another `Schedule` being added in some way to `self`
+ - max(|N<sub>C</sub>|) : the maximum number of instructions on any single channel within `self`
+
+### New Schedule methods
+
+| Method        | Runtime        |  Description     |
+|---------------|----------------|------------------|
+|`barrier(self)`| O(\|C\|) | Add `Delay`s across `channels` so that all of the `channels` are occupied until `self.duration(channels)`. |
```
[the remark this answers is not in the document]
lcapelluto: it has to be tracked

0000-qiskit-pulse-schedule-relative-timing.md
lcapelluto: I am thinking yes. even here, I think "do what I say" is the better policy to follow over "do what I mean" given the low level of control the user has over the instruments

0000-qiskit-pulse-schedule-relative-timing.md
lcapelluto: yes. it's not great, since delays are supposed to be blocking, but I don't think there's any other way to do this

0000-qiskit-pulse-schedule-relative-timing.md
lcapelluto: agree

itoko:
I realized we are using "relative time/timing" in the two meanings: "timing by channels (by-channels)" and "no t=0", i.e. "absolute time/timing" means "timing across channels (across-channels)" and "common t=0".
For example, 
```
`Instruction`s will not be played at an explicit time; rather, they will begin immediately after the last is done. 
```
is talking about "timing by channels", while
```
explicit timing of instructions will become confusing with the introduction of control-flow to quantum programs as it would become difficult to program experiments containing loops that are not unrollable.
```
is talking about "no t=0".

Regarding "by-/across-channels", 
I think we have two options:
- Build and save with "time counting by channels" (as of the current RFC)
- Build with "across-channels" and save with "by-channels"

I guess we are heading to "by-channels" because real devices do not support communication among channels. That makes "single time counting across channels" too much for real device.
So I don't disagree with heading "timing by channels" for Schedule (as a data class to be passed to backend) if communication among channels would never be supported, but I think we need to be careful when thinking about schedule builder (as a class for users to build pulse schedule). At the time building schedules, we often want to do `insert(time, instruction)`. So we may need to have intermediate (mutable) class to build schedule supporting `insert(time, instruction)`.
IMO, currently Schedule class is doing two things, building and storing, as like QuantumCircuit. While QuantumCircuit perform well because it's mutable, it is difficult for Schedule to perform well because it's immutable. I thought that's why we're going to introduce "builder class" for Schedule as like "StringBuilder" and "String" class in Java. Am I misunderstanding schedule builder?

Regarding "no/common t=0" considering control-flow, 
we can think of another option (I don't say better option).
What if we introduce a kind of "(control-flow) block" and "no t=0" among blocks but having "common t=0" within a block?

0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -52,49 +52,107 @@ Factors to consider:
...
 ### New Schedule methods
 
 | Method        | Runtime        |  Description     |
 |---------------|----------------|------------------|
-|`barrier(self)`| O(\|C\|) | Add `Delay`s across `channels` so that all of the `channels` are occupied until `self.duration(channels)`. |
-|`left_align(self)` | O(\|C\|) | This is the default behavior. We could have it return `self`, or we could remove any pre-pended `Delay`s. The latter is probably better, because `schedule.right_align().left_align()` should yield the original `schedule`. Should we also append `Delay`s? Likely. *Needs more thought* |
-|`center_align(self)` | O(\|C\|) | Pre- and ap-pend equally sized `Delay`s (how to split odd durations?) on each channel so that all channels are occupied until `self.duration()` |
-|`right_align(self)` | O(\|C\|) | Prepend `Delay`s on each channel so that all channels are occupied until `self.duration()`. Should we also remove trailing `Delay`s? |
-
+|`barrier(self, channels=None)`| O(\|C\|) | Add `Delay`s across `channels` so that all of the `channels` are occupied until `self.duration(channels)`. By default, barrier on all channels. |
```
taalexander: I believe this should be implemented as a series of `transforms` on schedules. For example, see my WIP [builder branch](https://github.com/taalexander/qiskit-terra/blob/pulse-builder-interface/qiskit/pulse/transforms.py). We do not want to build too many manipulations directly into a Schedule as this is not extensible. We should only build in what is necessary to allow these tools to be built. 

The builder interface will handle unifying user-level programming and the Schedule IR. I think the Schedule IR should concern itself with being a data structure with timing information and *some* helper methods for common pass operations.

0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -52,49 +52,107 @@ Factors to consider:
...
 ### API Changes
 
-\# TODO
-
 | Method        | Runtime        |  Change          |
 |---------------|----------------|------------------|
-| `append(self, schedule)` | O(\|N<sub>new</sub>\|) |  |
-|`insert(self, time, schedule)`| O(\|N<sub>new</sub>\| * \|N<sub>C</sub>\|) |  |
+| `append(self, schedule)` | O(\|C\|) | Previously, `append` would add `schedule` to `self` with relative times preserved, at time set by `self.duration({schedule.channels}.intersection({self.channels}))`. The new method will simply extend the instructions in `self` with those in `schedule` on a per channel basis. |
+|`insert(self, time, schedule)`| O(N<sub>new</sub> * N<sub>C</sub>) | This will be slow generally, but it can be implemented with the same runtime as `append` when the `time` is greater than `self.duration(schedule.channels)`. |
```
taalexander: Perhaps we should have both an `insert_at_time` and an `insert` (as the actual argument position in the schedule data structure) to properly differentiate these two operations?

0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -52,49 +52,107 @@ Factors to consider:
...
- - How new and existing methods will be updated
+The `Schedule` will have an internal attribute, `_insts_by_chan`:
+
+```
+_insts_by_chan: Dict[Channel, List[ScheduleNode]]
+```
+
+**Example**
+```
+# sched._insts_by_chan
```
taalexander: This doesn't necessarily work as many instructions are on multiple channels such as a barrier, or an `Acquire`. The internal data structure should be a list of `instructions`. It should be possible to request the times of these instructions which would dynamically evaluate them or pull from a cached value if this had already been evaluated and not invalidated.

lcapelluto: barrier is not an instruction, acquire operates over time on AcquireChannel

0000-qiskit-pulse-schedule-relative-timing.md
lcapelluto: ok, I'll remove these both because of this comment and as per discussion today

lcapelluto:
@itoko I don't understand your comment; I don't understand what you mean by "cross channels"

0000-qiskit-pulse-schedule-relative-timing.md
lcapelluto: by channels is better, and possible if you think of it as hardware channels

0000-qiskit-pulse-schedule-relative-timing.md
lcapelluto: sounds good to me

0000-qiskit-pulse-schedule-relative-timing.md
ajavadia: Yeah I think even if you have an instruction on multiple channels, you still can maintain a `_insts_by_chan` datastructure. You just repeat the instruction on both channels.

I mostly think about circuit schedules now, and in that case, given a 600 ns CNOT that engages 2 qubits, you would add a 600 ns Instruction to the timeline of both qubits.

0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -0,0 +1,103 @@
...
+| **RFC #**         | TBD                                          |
+| **Authors**       | Lauren Capelluto                             |
+| **Submitted**     | 3/11/20                                      |
+| **Updated**       | ---                                          |
+
+
+## Summary
+Update the internal structure of `Schedule`s to track instructions sequentially. `Instruction`s will not be played at an explicit time; rather, they will begin immediately after the last is done. To schedule idle time between instructions, `Delay` instructions must be used.
+
+The user API will be updated to take advantage of this relative timing, with several new methods: `barrier`, `left_align`, `center_align`, and `right_align` (see: [new methods](#new-schedule-methods)).
```
ajavadia, line 14:
```suggestion
The user API will be updated to take advantage of this relative timing, with several new methods: `barrier`, `left_align`, `center_align`, and `right_align`.
```
new methods section is delegated to the builder now.

0000-qiskit-pulse-schedule-relative-timing.md
```diff
@@ -0,0 +1,103 @@
...
+        ('instructions'=[Delay(duration=10, ...),
+                         Play(duration=100, ...),
+                         Play(duration=20, ...)],
+         'duration'=130),
+     DriveChannel(1):
+        ('instructions'=[Delay(duration=110, ...),
+                         Play(duration=20, ...)],
+         'duration'=130),
+    ...
+}
```
ajavadia, line 72: This seems to be the main takeaway of this RFC, so we should settle on it.

So the 130 number is here for performance right? Even though it could be computed by summing all durations on that channel. I'm good with that.

However, we should think about what a conditional will look like. If I don't know an instruction's delay, due to a branch or conditional, how does this datastructure work?

ajavadia:
I'm not totally clear on this justification. Can you clarify the complexity of `union` with absolute and relative datastructures? I don't think it's different with the currently proposed structure, but I could be wrong.

> It also encourages inefficient operations such as taking the union of two schedules. At first glance this operation should be efficient since all one needs to do is specifying the starting time of the instructions, but this requires verification that no instructions overlap in time in the two schedules resulting in expensive checks. The requirement for such a check to merge two schedules is obvious within the relatively schedules model, as one must look to make sure a delay exists in the first schedule for all instructions of the second schedule to merge.

0000-qiskit-pulse-schedule-relative-timing.md
lcapelluto: A `Schedule` is a "basic block", so conditionals should live on top of this. An alt approach (which I should list in this RFC) would be to make this explicit and allow absolute timing within a basic block even when we have control flow. I find that misleading (too "special case"-y), but, at the heart of it, they're equivalent.

lcapelluto: I think Thomas' new RFC touches on this

ajavadia:
Can we list the types of operations that we do in schedules (append, insert, delete/replace, union, find channel idle times) and the complexity of those in a relative vs. absolute data structure? I think for the basic blocks we should choose what makes compiler manipulation faster. I agree they can be converted to each other, but their manipulation is not the same.

lcapelluto:
@ajavadia I agree. I'll do that 👍

ajavadia:
Ok so it based on this, `append` and `shift` are faster in relative, `insert` faster in absolute?

If I compare this with #18 it seems the findings are different in the case of `insert`. Can we come to an agreement on this @lcapelluto @itoko? Because I think the timed circuit and pulse case should be pretty identical in terms of datastructure and manipulation runtime.

In any case I just wanted to make sure we are basing this on some concrete runtime analysis, which will also help in the implementation. I think it is clear that relative wins in terms of the most common operation which is append.

PR #18, opened 2020-04-16
[WIP] [RFC] Scheduled circuit

itoko:
Add "scheduled circuit" in-between “gate circuit”(QuantumCircuit) and “pulse schedule”(pulse.Schedule) to enable time-aware circuit-based noise simulation.

### DIFF
--- a/0000-scheduled-circuit.md
+++ b/0000-scheduled-circuit.md
@@ -24,7 +24,7 @@ as well as scheduling methods to construct it from a QuantumCircuit.
 
 ## User Benefit
 - Time-aware circuit-based noise simulation on Aer is enabled.
-- Scheduling in circuit level will enable more optimization than that in pulse level in the future.
+- Scheduling at the circuit level will enable more optimization than that of the pulse level as the semantic meaning of the circuit gate is still available whereas at the pulse level this is lost.
 
 ## Design Proposal
 ### High-level API:
--- a/0000-scheduled-circuit.md
+++ b/0000-scheduled-circuit.md
@@ -29,7 +29,7 @@ as well as scheduling methods to construct it from a QuantumCircuit.
 ## Design Proposal
 ### High-level API:
 1. Noise simulation flow (new):
-```
+```python
 timed_circuit = circuit.schedule(transpiled_circuit, ...)  # transpiled_circuit: QuantumCircuit
 qobj = assemble(timed_circuit)  # timed_circuit: ScheduledCircuit
 ```
--- a/0000-scheduled-circuit.md
+++ b/0000-scheduled-circuit.md
@@ -1,4 +1,4 @@
-# WIP: [RFC] Scheduled Circuit
+# [WIP] [RFC] Scheduled Circuit
 
 | **Status**        | **Proposed/Accepted/Deprecated** |
 |:------------------|:---------------------------------------------|
@@ -230,7 +230,7 @@ _timed_inst_by_qubit: Dict[Qubit, List[TimedInstruction]] # lists are sorted by 
   - O(N): update start times
   - O(ListReplace(D)*W): replace items on lists by qubit
 
-In addition to these, we need kinds of “find()” family for users to specify refs in insert/replace etc.
+In addition to these, we need kinds of “find()” family for users to specify references in insert/replace etc.
 
 \*) I assume the interface that references are given as Postions. If references should be given as Instructions, we need to accept overhead to find positions. 
 Such overhead would be O(D*W) for Relatively-timed and O(log(D)*W) for Absolutely-timed unless we add another data structure for the purpose.
--- a/0000-scheduled-circuit.md
+++ b/0000-scheduled-circuit.md
@@ -64,7 +64,7 @@ Technical reference level design. Elaborate on details such as:
 - Reference definition, eg., formal definitions.
 
 ### Example code
-```
+```python
 def schedule(qc: QuantumCircuit,
              backend_props: BackendProperties,
              method: str = "alap",
--- a/0000-scheduled-circuit.md
+++ b/0000-scheduled-circuit.md
@@ -14,7 +14,7 @@ to make time-aware circuit-based noise simulation possible.
 
 ## Motivation
 Currently there is no way to instruct `delay` (or `wait`/`idle`) for a specific duration in circuit level.
-This prevents from easy noise simulation of experiments with `delay`s, e.g. T1/T2 measurement, on Aer.
+This makes simulations with time-dependent noise, e.g. T1/T2 measurement, difficult on Aer.
 A new circuit class for incorporating with instructions having `duration` should be added,
 as well as scheduling methods to construct it from a QuantumCircuit.
 
--- a/0000-scheduled-circuit.md
+++ b/0000-scheduled-circuit.md
@@ -1,0 +1,250 @@
+# WIP: [RFC] Scheduled Circuit
+
+| **Status**        | **Proposed/Accepted/Deprecated** |
+|:------------------|:---------------------------------------------|
+| **RFC #**         | ####                                         |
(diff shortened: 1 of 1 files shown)


0000-scheduled-circuit.md
```diff
@@ -0,0 +1,266 @@
...
+```
+
+2. Circuit-to-Schedule flow (update):
+```
+schedule = scheduler.circuit_schedule(transpiled_circuit, ...)  # transpiled_circuit: QuantumCircuit
+```
+->
+```
+timed_circuit = circuit.schedule(transpiled_circuit, ...)  # transpiled_circuit: QuantumCircuit
+schedule = pulse.build_schedule(timed_circuit, ...)  # timed_circuit: ScheduledCircuit
```
taalexander, line 44:
```suggestion
schedule = pulse.lower_circuit(timed_circuit, ...)  # timed_circuit: ScheduledCircuit
```

How do you plan to establishing timing for circuits in the case where an instruction blocks resources it is not directly operating on at a semantic level? For example a cross talk compensation pulse inserted for a neighboring qubit. To do this properly likely requires the notion of both target (operated) and resource acquired qubits. If you were to schedule this at the circuit level and then lower to the pulse level (`build_schedule`) it would result in two different programs being executed with different timings. I think the desired behavior is for the timing to be equal.

0000-scheduled-circuit.md
```diff
@@ -0,0 +1,266 @@
...
+## Detailed Design
+Technical reference level design. Elaborate on details such as:
+- Implementation procedure
+  - If spans multiple projects cover these parts individually
+- Interaction with other features
+- Dissecting corner cases
+- Reference definition, eg., formal definitions.
+
+### Example code
+```
```
taalexander, line 67:
```suggestion
```python
```

0000-scheduled-circuit.md
```diff
@@ -0,0 +1,266 @@
...
+timed_circuit = circuit.schedule(transpiled_circuit, ...)  # transpiled_circuit: QuantumCircuit
+qobj = assemble(timed_circuit)  # timed_circuit: ScheduledCircuit
+```
+
+2. Circuit-to-Schedule flow (update):
+```
+schedule = scheduler.circuit_schedule(transpiled_circuit, ...)  # transpiled_circuit: QuantumCircuit
+```
+->
+```
```
taalexander, line 42:
```suggestion
```python
```

0000-scheduled-circuit.md
```diff
@@ -0,0 +1,266 @@
...
+2) insert gates in idle timeslots
+
+## User Benefit
+- Time-aware circuit-based noise simulation on Aer is enabled.
+- Scheduling in circuit level will enable more optimization than that in pulse level in the future.
+
+## Design Proposal
+### High-level API:
+1. Noise simulation flow (new):
+```
```
taalexander, line 32:
```suggestion
```python
```

0000-scheduled-circuit.md
```diff
@@ -0,0 +1,266 @@
...
+def schedule(qc: QuantumCircuit,
+             backend_props: BackendProperties,
+             method: str = "alap",
+             dependency: Optional[DependencyGraph] = None) -> ScheduledCircuit:
+    """
+    Args:
+        qc: a transpiled circuit to be scheduled.
+        backend_props: to get gate_length
+        method: same as in pulse.scheduler.schedule_circuit
+        dependency: (for future) asap and alap can work without this
```
taalexander, line 77: What is this for?

0000-scheduled-circuit.md
```diff
@@ -0,0 +1,266 @@
...
+    """
+    if dependency is None:
+        # schedule based on the standard dependency that is the same as used in DAGCircuit
+        # algorithm is almost the same as in pulse.scheduler.methods.algorithm
+    else:
+        # scheudle based on the extended dependency for the future
+
+
+Qubit = int # assume only physical qubits
+ArgumentedInstruction = Tuple[Instruction, List[Qubit], List[Clbit]]
```
taalexander, line 89:
```suggestion
ArgumentedInstruction = Tuple[Instruction, List[Qubit], List[Clbit]]
```

Would it be possible for this to become standard for operands to be accepted in the instruction definition? Having a tuple format is really limiting in the long run. It makes it harder for us to standardize on an IR between circuit and pulse. For example, what happens if a circuit call out to a function or equivalently jumpts to a label in the program. This tuple format doesn't provide a location for that function/label to go.

If alternatively 
```python
class Instruction(abc.ABC)

def __init__(self, operands, metadata)

class CircuitInstruction
     def __init__(self, qubits, clbits):
          super().__init__([qubits, clbits])
```
This gives us a path forward to IR unification.

0000-scheduled-circuit.md
```diff
@@ -0,0 +1,266 @@
...
+                   ...) -> pulse.Schedule:
+    """Build schedule by replacing instructions in timed_circuit with schedules defined in inst_map.
+    """
+```
+
+## Alternative Approaches
+### float time vs int time:
+- float (seconds) = gate time in BackendProperties
+- int (number of cycle time) = pulse duration is defined by samples with dt or dtm
+  - -> float time for now (convert float time into int time in pulse.build_schedule())
```
taalexander, line 153: I do not think floating-point is a good idea. This will likely result in many difficulties in the actual implementation as there will be floating-point comparisons everywhere which will have many subtleties and be costly from a computational perspective. This is asking for many many headaches of having to check `assert_almost_equal` etc. when scheduling. With an integer-based formulation this calculation will have to be done only once, without comparison in Aer for the calculation of dissipators.

Furthermore, it will make moving towards the notion of a cycle much more difficult which is very crucial from a long term perspective as we introduce control flow. This also will increase the non-uniformity between pulse and circuits, when we should be trying to move towards more uniformity.

0000-scheduled-circuit.md
```diff
@@ -0,0 +1,266 @@
...
+- Dictionary (Dict[Qubit, List[LengthedInstruction]] discussed above)
+- DAG (as like used in DAGCircuit)
+
+Note: *For any node of DAGCircuit, its in-(or out-)edges must have different qubit label.
+It enables DAGCircuit to store the ordering of instruction by qubits.*
+As for general DependencyDAG, which represents dependencies among instructions,
+its node can have multiple in-(or out-)-edges with the same qubit label.
+So DependencyDAG is not appropriate to store the ordering of instruction by qubits.
+
+In Dictionary, any multi-qubit instruction is stored at multiple positions on multiple lists.
```
taalexander, line 248: For this reason I do not believe a dictionary is very viable, maintaining these qubit dictionaries I think is equivalent to maintaining a list of integers for each qubit, where the integer specifies the index of the instruction in the `instructions` list. These must be updated on the append/removal of an instruction making these operations O(j*k) where  is the average number of qubits and k the average number of instructions per qubit. This could be slightly mitigated if instructions could be hashed, which may not be true in general.

Consequently I think the two clear choices for an IR structure (as in most IRs) are the linear IR (a list of instructions) or a DAG. I think the base IR makes sense to be linear as it is much simpler and if necessary we can support the DAG version via a conversion.

0000-scheduled-circuit.md
```diff
@@ -0,0 +1,266 @@
...
+So referring a multi-qubit instruction with one position is not possible.
+In DAG, it is possible because any node can be specified by "instruction id (an integer)".
+To get instruction by its id, we may need to store a map from id to instruction (as list or dict).
+
+
+
+## Questions
+- any other use case? noise simulation with custom gate (with custom duration and noise)?
+
+- need to support "insert_at_time()"?
```
taalexander, line 258: I believe so. Handling overlaps with floating point time will be more nuanced and error prone.

PR #19, opened 2020-05-01
[RFC][WIP] A more scalable pulse visualizer

nkanazawa1989:
This proposes replacement of current pulse drawer based on our experience of writing Qiskit Pulse paper. User will enjoy benefit of improved programming efficiency and high quality visualization for publication.

### DIFF
--- a/0000-new_pulse_drawer.md
+++ b/0000-new_pulse_drawer.md
@@ -34,7 +34,7 @@ Usually a pi pulse duration is around 160 cycles whereas that of the measurement
 Currently, we can truncate the measurement pulse with the `plot_range` option.
 However, when we enable conditional gates or a conditional reset in the middle in the schedule, the truncation by `plot_range` doen't work becase users may also want to know the pulse sequence after the measurement stimulus.
 There is another frustrating situation based on our experience. When we wrote the Qiskit Pulse paper we needed to modify the drawer itself to address reviewer's requests and we decided not to share the code for pulse drawing in the public database.
-To enrich the style sheet option will help us in such situation because we just need to create a dedicated style sheet for publication and share the style sheet instead of the whole modified Qiskit-terra.
+Having configuration be based on a style sheet will help us in such a situation because we would just need to create a dedicated style sheet for publication and share the style sheet instead of modifying source code.
 
 From the point of view of contoributors, current drawer is implemented as one huge callback function and it is very hard to maintain.
 Sometime a change in Qiskit Pulse module may cause an unexpected error in the drawer or even crash it, but this will not be detected because visualization level test is skipped now.
--- a/0000-new_pulse_drawer.md
+++ b/0000-new_pulse_drawer.md
@@ -44,7 +44,7 @@ Thus, it is better to decompose this huge callback function into some unit funct
 This update will improve the efficiency of pulse programming and maintainability of code.
 
 ## User Benefit
-Qiskit Pulse user will enjoy the benefit. From novice to professional level, users will be able to get a readable visualization with more simple function or method call for debugging purpose and then easily customize the output format for publication.
+Every Qiskit Pulse user will enjoy the benefit. From novice to the professional level, users will be able to get a readable visualization with a more simple function or method call for debugging purposes and then easily customize the output format for publication.
 Cost of code maintainance will be also decreased.
 
 ## Design Proposal
--- a/0000-new_pulse_drawer.md
+++ b/0000-new_pulse_drawer.md
@@ -32,7 +32,7 @@ Another defect can be seen when we try to visualize two pulses whose durations a
 A good example is a single qubit pi pulse followed by a measurement stimulus pulse.
 Usually a pi pulse duration is around 160 cycles whereas that of the measurement pulse is over 10000.
 Currently, we can truncate the measurement pulse with the `plot_range` option.
-However when we enable conditional gates or a conditional reset in the middle in the schedule, the truncation by `plot_range` doen't work becase users may also want to know the pulse sequence after the measurement stimulus.
+However, when we enable conditional gates or a conditional reset in the middle in the schedule, the truncation by `plot_range` doen't work becase users may also want to know the pulse sequence after the measurement stimulus.
 There is another frustrating situation based on our experience. When we wrote the Qiskit Pulse paper we needed to modify the drawer itself to adress reviewers requests and we decided not to share the code for pulse drawing in the public database.
 To enrich the style sheet option will help us in such situation because we just need to create a dedicated style sheet for publication and share the style sheet instead of the whole modified Qiskit-terra.
 
--- a/0000-new_pulse_drawer.md
+++ b/0000-new_pulse_drawer.md
@@ -161,8 +161,14 @@ qiskit
       +-- drawing_engine.py (collection of small programs to help object mapping)
 ```
 The drawer function `schedule_drawer` and `sample_pulse_drawer` creates a static 2D image of waveforms.
+In the existing drawer, those interfaces are integrated as the `pulse_drawer` and this function internally calls corresponding drawer class.
+However, there are many unused arguments in the interface when `SamplePulse` is drawn because `Schedule` visualization is more complicated and requires some additional options.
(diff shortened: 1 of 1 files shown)


0000-new_pulse_drawer.md
```diff
@@ -0,0 +1,274 @@
...
+The interval between `Play` instructions should be padded by `Delay` and this appears as long zero array with the length of the interval in terms of system cycle time (not actual pulse time).
+This is because we need to truncate long pulses or long delays based on `truncate` preference in the style sheet.
+If we output the entire waveform as a single data array it is very hard to recover a length of each pulse.
+The method `max_amp` and `min_amp` are used to determine the autoscaling value of the channel.
+
+First the `schedule_drawer_core` initialize `EventsOutputChannels` for all channels specified by the user input and the style sheet preference and keep them as a python `dict`. Then, those channels are rearanged as a `list` according to the `layout` option in the style sheet. If `'layout': 'channel-type-wise'` is specified, the program collects all `DriveChannel`s in the dictionary and then collects `ControlChannel`s, ... , so that channels of the same type are shown in one group.
+
+In `utils.py` several helper functions are defined to extract information and coordinate to show from the `EventsOutputChannels`.
```
taalexander, line 208: Does this need to be in a utils module? Could it just live in one of the other modules?

0000-new_pulse_drawer.md
```diff
@@ -0,0 +1,274 @@
...
+This is a big change from current implementation (but not impact to users) such that existing huge callback function is decomposed into small pieces which can be unittested.
```
taalexander, line 209:
```suggestion
This is a big change from the current implementation (but not impactful to users) such that existing huge callback function is decomposed into small pieces which can be unittested.
```

0000-new_pulse_drawer.md
```diff
@@ -0,0 +1,274 @@
...
+For example,
+
+```python
+def get_labels(manager: EventsOutputChannels) -> List[Tuple[Tuple[float, float], str]]:
+  # return coordinate and name of each pulse instruction
+  pass
+```
+
+this is the callback function to extract a set of pulse names (sometime formatting names by request) and the coordinate to draw of each pulse in the `EventsOutputChannels`.
```
taalexander, line 218: These are just functions, not `callback function`. A [callback](https://en.wikipedia.org/wiki/Callback_(computer_programming)) has a specific meaning.

0000-new_pulse_drawer.md
```diff
@@ -0,0 +1,274 @@
...
++ `get_lengths`: return length of each pulse
++ `get_phases`: return phase change/set instruction
++ `get_frequencies`: return frequency change/set instruction
++ `get_snapshots`: return snapshot instruction
++ `scale_waveform`: apply scaling to pulses
++ `truncate_waveform`: truncate long waveform and delay
+
+We can prepare unittest for those functions as well.
+
+### Phase3: Add style sheets
```
taalexander, line 251: I am unclear how the stylesheets will be implemented and I believe this has an impact on the implementation discussed above. From my understanding, it seems you wish these to be truly configurable. There really are two options, they define fixed-configuration settings and the code above reads these and does what it will. Or the stylesheet provides functionality implementations via dependency injection which is more work to implement but more flexible in the long run. When you mention regexes that is what this evoked in my mind but it is unclear.

0000-new_pulse_drawer.md
```diff
@@ -0,0 +1,274 @@
...
+### Phase4: Add core function and interface
+With all of above PRs we can start implementing main part of drawer.
+The unittest for `matplotlib.Figure` output is skipped in the current testing framework.
+We need to check the output of figure by our eyes.
+
+## Alternative Approaches
+N/A
+
+## Questions
++ How old drawer can be deprecated? Becase interface will change drastically, I feel introducing new drawer independently to old one is easier.
```
taalexander, line 268: So long as calling it without arguments is still possible we can replace the old drawer and break code for such a big change. We can introduce the new drawer slowly with a different function name such as `drawv2` to test over time.

0000-new_pulse_drawer.md
nkanazawa1989: You're right. The design of style sheet will influence the implementation of functions. I'll update this part.

> There really are two options,

I agree with the second option gives us a huge degree of freedom, but I think the first option is more preferable because of two reasons.

1. security risk
I assume style sheets are serializable so that they can be easily shared among external communities. Thus code injection is not considered and all dependencies will be packed into `drawing_engine.py` (previously I call it `utils.py`) because they are tested and the code is trustable.
In CSS community, some security risks are reported due to code injection.
https://www.netsparker.com/blog/web-security/private-data-stolen-exploiting-css-injection/

2. maintainability
I experienced pickled object is very sensitive to version change and they are prone to be broken by major update. I think style sheets should be stable over code changes. This discussion realized me to add `to_dict` and `from_dict` method to base style sheet class.

nkanazawa1989:
@taalexander Thanks for your kind reviewing. It was actually full of suggestions. I like the idea of dynamic drawer and we should mention it in the RFC. My thought for the dynamics drawer is something like
```
sched = dynamic_drawer(sched)
dynamic_drawer.draw()

sched += Play(my_pulse, d0) # my_pulse appears in the canvas
sched += Play(measure, m0) # measure pulse appears in the canvas
```
and this is very much useful for educational purpose and hands-on. Currently schedule is designed as immutable object and I have no idea how we can easily track such object, but I agree at least drawer core should be able to support this workflow.

Considering this situation, the core object should be a class rather than function. I'll update the design of the core.

nkanazawa1989:
The design of core functions are replaced with classes so that we can replace drawing backend. I think it becomes clear how we can reuse the core for creating the dynamic drawer.

0000-new_pulse_drawer.md
taalexander: 1. I don't see a reason that they have to be serializable, this seems like a lot of effort to guarantee for little payoff. Consider [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)
2. This same point seems to consider you're serializing it?

Regardless I think we are proposing two separate and mutually inclusive ways of configuration. The second way I was proposing is a dependency injection pattern and would lead to more customizable and extendable plotters.

taalexander:
> @taalexander Thanks for your kind reviewing. It was actually full of suggestions. I like the idea of dynamic drawer and we should mention it in the RFC. My thought for the dynamics drawer is something like
> 
> ```
> sched = dynamic_drawer(sched)
> dynamic_drawer.draw()
> 
> sched += Play(my_pulse, d0) # my_pulse appears in the canvas
> sched += Play(measure, m0) # measure pulse appears in the canvas
> ```
> 
> and this is very much useful for educational purpose and hands-on. Currently `Schedule` is designed as an immutable object and I have no idea how easily we can track such object, but I agree at least drawer core should be able to support this workflow.
> 
> Considering this situation, the core object should be a class rather than function. I'll update the design of the core.

By dynamic drawer I meant one that was capable of scrolling, zooming in, or boxing and unboxing certain long pulses. Regardless, we are moving away from immutable Schedules as they have served their purpose but we need to transition to a more fully-featured IR.

0000-new_pulse_drawer.md
nkanazawa1989: I think the value of stylesheets is the sharability so that everyone can easily reproduce one's work, or we can save favorite custom config in the file and save the time for writing code in every notebook. So serialization is good approach because serialized data has a little possibility to execute malicious code. And injected code, or callback function, cannot be saved as a file. Maintainability another reason.

I agree the code injection can extend the ability of plotter, but for me it sounds like add-on or pulgin software. How do you think the below example?

```python
.draw(style=my_favorite_style, add_on=[format_pulse_name, add_scroll_bar])
```

In this way we can share the stylesheet with colleagues while keeping extendability of the plotter by code injection.

Maybe I don't understand the _injection pattern_ correctly.

nkanazawa1989:
I'm used to mention it as interactive. I think dynamic or interactive module can be implemented by using matplotlib with [tkinter](https://docs.python.org/3/library/tkinter.html). I think code injection to the stylesheet is reasonable approach when such controllability on the figure is required. I need to make a rough design of interactive so that we can get proper plotter design with sufficient extendability.

PR #20, opened 2020-05-14
[RFC] BackendConfiguration.supported_instructions

1ucian0:
Backends support different set of instructions. We need to improve that.

### DIFF
--- a/####-supported_instructions.md
+++ b/####-supported_instructions.md
@@ -1,4 +1,4 @@
-# BackendConfiguration.supported_instruction
+# BackendConfiguration.supported_instructions
 
 | **Status**        | **Proposed** |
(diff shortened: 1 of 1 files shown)


taalexander:
This is part of a much larger class of information the Qiskit compiler will require. I think whatever is done here we should consider how we should supply this and where this information should go. Otherwise, we will end up with random keys being supplied by backends with no structure. 

In the long run we will run into many of the problems addressed in the [LLVM target machine](https://llvm.org/docs/WritingAnLLVMBackend.html#id8). I am not arguing we try and solve this problem in one fell swoop but we should make sure that we are aware of the complexity of the problem and not back ourselves into a corner.

1ucian0:
I agree with @mtreinish. Closing.

PR #24, opened 2022-01-26
Create ####-rfc-unitary-subroutines.md

ndebeaudrap:
A proposal for a new variety of subroutine, which is in a sense intermediate between gate and def subroutines, providing some of the benefit of each as a means to describe more complex quantum procedures, drafted in response to the discussions of the Circuits and Generics WG.


1ucian0:
It seems this RFC is a proposal for a change in OpenQASM specification instead of Qiskit. @ndebeaudrap should consider opening an issue in https://github.com/openqasm/openqasm/issues with this proposal.

PR #25, opened 2022-05-31
`Operation`: interface for `QuantumCircuit` operations

jakelishman:
I'm hoping to give us a place to keep notes from meetings, and a place to discuss asynchronously/offline on this proposed new interface.  For the first pass, I've tried to write up my understanding of what's been discussed so far, but feel free to challenge my interpretations on those - that's what this is here for.

Other people who have been very involved in this discussion previously:

- @mtreinish
- @kdk
- @alexanderivrii
- @eliarbel

and some people who I think are also invested and may want to comment on the process:

- @ShellyGarion
- @ajavadia
- @ikkoham
- @1ucian0

### DIFF
--- a/0025-interface-for-circuit-operations.md
+++ b/0025-interface-for-circuit-operations.md
@@ -1,0 +1,181 @@
+# `Operation`: the interface for valid `QuantumCircuit` operations
+
+| **Status**        | **Proposed**                                 |
+|:------------------|:---------------------------------------------|
+| **RFC #**         | ####                                         |
+| **Authors**       | Jake Lishman (jake.lishman@ibm.com)          |
+| **Deprecates**    | None                                         |
+| **Submitted**     | 2022-05-31                                   |
+| **Updated**       | 2022-05-31                                   |
+
+
+## Summary
+
+Currently, only subclasses of `Instruction` can be put on `QuantumCircuit`, but
+this interface has become unwieldy and includes too many methods and attributes
+for general-purpose objects.  We propose to put a new interface in place that is
+far more stripped down, and to remove transpilation-specific methods from the
+objects themselves, and move them to make them transpiler behaviour.  We propose
+that this new interface will be called `Operation`.
+
+
+## Motivation
+
+- `Instruction` as a complete type has too many unrelated attributes and methods
+  from the base interface, which not all "quantum operations" can reasonably
+  fulfil.  For example, `definition` (which is already circumvented in several
+  cases), `duration` and `unit` (which only have meaning after transpilation),
+  and `assemble` (which supposes that the object is immediately executable).
+
+  Some of these are fine _after transpilation_, but are too restrictive for
+  `Instruction` to be the only input type for the transpiler.
+
+- `Instruction` is defined as part of `qiskit.circuit`, which poses cyclic
+  dependency issues for some lower-level objects (e.g. those based in
+  `quantum_info` types).
+
+- Having a simpler, abstract interface (`Operation`) allows us to break the
+  cyclical dependencies (it can be the lowest part of Terra), make more objects
+  suitable for being "transpiler input", and reduce the memory/implementation
+  complexity for those objects that can be.
+
+- `Instruction` need not be replaced by this; we still want something to
+  represent "a transpiled operation".  This is a greater restriction than "an
+  operation that can be transpiled", so `Instruction` can (and will) also
+  implement `Operation`.
+
+- Defining an immutable interface lets us re-use the same objects more during
+  transpilation, which will save a lot of memory.  For example, we will not need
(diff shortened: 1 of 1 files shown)


alexanderivrii:
@jakelishman, many thanks for taking the discussion on the `Operation interface` to a completely new level. We have been also thinking about this with @eliarbel and @ShellyGarion, and here are some additional thoughts and questions.

We agree that we do not want `Operations` to carry around their `definitions`. Does it mean that we also want to remove `definitions` for all of the already existing gates (including the basic gates, such as Hadamards)? We can probably store these `definitions` in the `equivalence library`, if they are not there already.

Regarding the transpilation process. I am in favor of having a special transpiler pass for each `Operation` type, very similarly to `UnitarySynthesis` transpiler pass for unitaries. So, we would have a `LinearSynthesis` transpiler pass (we have it already, but it does not support parameters such as `coupling_map` and is not `Target`-aware), `CliffordSynthesis`, `MCXSynthesis`, and so on. Or maybe we can have a single transpiler pass that can take care of all of the different `Operations`, but such that we can also specify which types of `Operations` to synthesize and which not to. (Of course, at the end we want the transpiler to synthesize all `Operations`, but when making intermediate decisions we may only want to synthesize a subset of the types).  I also like the `plugin` capability offered by `UnitarySynthesis` and think that it would also benefit these special transpiler passes. This ties well with @jakelishman's point:

> Researchers will be able to more easily "plug in" their own synthesis methods for higher-order objects at transpilation run time.

The suggested interface also means that `Operations` are not required to have a `to_operator` or `to_matrix` methods, I don't know if this might become problematic.

To better understand the difference between `parameter_spec` and `state`: something like rotation angles for rotation gates are considered `parameters`, while something like `destabilizer/stabilizer` table for `Cliffords` would be considered `state`. Is this understanding correct?

Implementation-wise, the proof-of-concept 7966 on adding `Cliffords` to `QuantumCircuits` also required `Operations` to support broadcasting, `condition` and `_directive`, none of which we want in the API. So we will need to handle each of these somehow in the clean implementation of the `Operation` class.

0025-interface-for-circuit-operations.md
```diff
@@ -0,0 +1,181 @@
...
+need to care about the circuit model, since they are using a different
+formalism.  This also means that objects with more than one synthesis method
+have to define a "preferred" synthesis method.  This is an issue for things like
+the multi-controlled X.
+
+
+## Questions
+
+1. At this point, the exact requirements for `Operation` are not fully
+   finalised.  This RFC will not be ready for implementation until they are.
```
mtreinish, line 153: I think we can implement this rfc partially to add the class and expand the interface over time as long as we do it a backwards compatible manner (which is what I believe the approach in Qiskit/qiskit-terra#7966)

```suggestion
1. At this point, the exact requirements for `Operation` are not fully
   finalized.  This RFC will not be ready for a complete implementation until they are.
```

0025-interface-for-circuit-operations.md
```diff
@@ -0,0 +1,181 @@
...
+
+2. Will classical operations also be `Operation`, or will we do something
+   different with them?
```
mtreinish, line 156: For me I think it makes the most sense to have classical operations be based on the same base class. We can specialize on top of it if there are stricter requirements, but having a common class for all operations makes working with circuits internally a lot easier I think. Is thing a operation, if so we can use it in a circuit? Then for things consuming circuits at a low level (like qpy) know how to work with the base element in the circuit.

0025-interface-for-circuit-operations.md
```diff
@@ -0,0 +1,181 @@
...
+
+3. What information about necessary qubits should we expose, bearing in mind
+   that the "number of qubits needed" may be dependent on the synthesis method?
```
mtreinish, line 159: I think just the number of qubits the operation acts on directly. If a synthesis method requires an ancilla, it's the responsibility of the synthesis method to know that check and either fallback to some other synthesis method or error appropriately if it can't work.

mtreinish (approved):
This LGTM, the proposal seems straightforward and matches up with my memory of discussions on this topic. Qiskit/qiskit-terra#7966 is starting the implementation here by adding the base class operation class and the required parameters of  name and num_qubits. After that merges we can work on expanding the class definition with the parameter spec (which I assume will involve #26) and the state (which I'll need to write a design doc for) fields. I'm not sure if we want to resolve the questions before merging this. At least questions 2 and 3 have clear answering to me.

jakelishman:
I'll merge and close this, given it's already landed in Terra.

jakelishman:
Whoops, pressed close by accident.  I don't have merge power on this repo, as it turns out.

PR #26, opened 2022-07-15
Add partial draft of run-time parameters RFC

jakelishman:
Apologies, there's quite a lot more I intended to write (and refine the current text) about further parts of the design and my progress, but I now need this to be up and available for others immediately.

I have pushed the branch [`move-instruction-parameters` to my fork](https://github.com/jakelishman/qiskit-terra/tree/move-instruction-parameters) which contains what I have currently done.  Roughly, this is:

- added most of the backwards compatibility shim for `Instruction.params`
- added a `parameters` field to `CircuitInstruction`, `DAGOpNode` and `DAGDepNode` (the last two might not be complete - I don't think I had fully finished updating `DAGCircuit` and `DAGDependency`)
- updated the `CircuitInstruction` context class to insert a reference to themselves into contained `Instruction` instances
- added a first pass at defining a "parameter specification" for `Instruction` instances
- added a first (poor) pass at a new `Instruction.decompose` interface to replace `Instruction.definition`.  What I've done isn't right at all - my failure here is what caused me to push this from Terra 0.21.  The starts of the design in the RFC about defining "input parameters" for `QuantumCircuit` are the right way forwards here, and the weird attempts currently in the branch to handle `Instruction.decompose` and `Instruction.definition` should be completely redone.
- made the initialisers of some gates accept zero parameters - in the new preferred form, the rz instruction will be initialised as `RZGate()`, and it's only when we add it to a circuit that we define its operands.

Certainly still to do:
- the "parameter specification" probably needs tweaking (I would keep everything about this private for Terra 0.22, because we'll want to refine the system heavily for typed dynamic parameters)
- all uses of `Instruction.params` (whether through circuits or DAGs) in Terra need updating to the new form. 
- `QuantumCircuit` needs to learn how to specify input parameters, so we can define the new interface for `Instruction.decompose` (or `Instruction.definition`).
- The new `Instruction.decompose` / `Instruction.definition` method needs to be in place so ad-hoc no-subclass `Instruction` instances can be created and parameterised correctly (e.g. the return value from `Instruction.inverse`, etc).

Please don't take any of this RFC / any of the code I wrote in my branch on Terra as gospel - a lot of the interfaces doubtless need tweaking or rewriting.

### DIFF
--- a/0026-gate-parameters-as-runtime-values.md
+++ b/0026-gate-parameters-as-runtime-values.md
@@ -1,0 +1,541 @@
+# Run-time classical calculations within circuits
+
+| **Status**        | **In development**                           |
+|:------------------|:---------------------------------------------|
+| **RFC #**         | ####                                         |
+| **Authors**       | Jake Lishman (jake.lishman@ibm.com)          |
+| **Submitted**     | 2022-07-15                                   |
+| **Updated**       | 2022-07-15                                   |
+
+
+## Summary
+
+To support dynamic circuits, we need to be able to represent general classical
+values and manipulation of these values within circuits.  Instructions and other
+gates should be able to be parameterised in terms of these run-time values.
+
+
+## Motivation
+
+The OpenQASM 2-like model of circuit execution previously in use does not allow
+any interaction with classical values during execution of a circuit, other than
+conditioning a single gate on whether a register has one exact numeric value.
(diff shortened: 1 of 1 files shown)


jakelishman:
This is long since out-of-date, and anything left is obsoleted by the new Rust-space circuit-data work. I'll close this to avoid the clutter.

PR #28, opened 2022-08-25
Dataframe for Qiskit Experiments

nkanazawa1989:
The complicated data structure inside the Qiskit Experiments class, particularly `ExperimentData`, makes advanced usage of the experiment framework harder. Experimentalists may hesitate to use such framework for their research project. Based on our users survey, we propose new experiment data structure built on top of the pandas dataframe, which is one of the most celebrated python package in the data science community (and we can also reduce maintenance overhead of custom build classes). 

This PR focuses on the update of analysis results. In future we can also adopts data frame for experiment (circuit) results, i.e. `ExperimentData.data()`, for better metadata management.


0000-experiment-dataframe.md
```diff
@@ -0,0 +1,190 @@
...
+  "01": 10,
+  "10": 34,
+  "11": 13,
+}
+```
+This dictionary can be represented as follows.
+```python
+cc_array = array([25, 10, 34, 13])
+```
+Note that, for example, the index `2` is `0b10` in the binary representation, and thus we can translate bit string "10" to index 2 of the array, given we know the size of memory slot. In the same way, for the CS data,
```
eggerdj, line 55: This seems unscalable to me. If I am working with 20 qubits then this implies an array with length `2**20`?

nkanazawa1989: We had the same discussion in one of past Qiskit demo days before. I agree this is not scalable, but we can use sparse array in that case because realistically we don't set shots of 2**20. Usually in QE, in the context of QCVV, we use small number of qubits so this doesn't matter for the most of cases.

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+- extra (Dict): Same as `AnalysisResultData.result_data["_extra"]`. Free-form dictionary for analysis metadata.
+- experiment_id (str): Same as `AnalysisResultData.experiment_id`. The ID of the associated experiment.
+- verified (bool): Same as `AnalysisResultData.verified`. A boolean value to indicate if this entry is checked by human experimentalist. This can be removed since this is rarely used.
+- tags (List[str]): Same as `AnalysisResultData.tags`. An arbitrary strings to help filtering this entry.
+- source (Dict): Same as `AnalysisResultData.result_data["_source"]`. Version information of QE that was used to generate this entry.
+- created_in_db (bool): Same as `AnalysisResultData._created_in_db`. A boolean value to indicate wether this is loaded from the database or just created locally.
+- result_id (str): Same as `AnalysisResultData.result_id`. A unique ID of this entry.
+- backend (str): Same as `AnalysisResultData.backend_name`. Name of the associated backend.
+- creation_datetime (datetime): Same as `AnalysisResultData.creation_datetime`. A date time at which the experiment data received the job results.
```
chriseclectic, line 76: Should we have a `modified` field as well, for the case where a result is modified (eg if a use sets tags, verified etc).

chriseclectic, line 76: It might also be good to keep these field names short, since they will be rows in pandas (eg `created` instead of `creation_datetime`)

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+
+### A. Experiment data
+
+We propose to replace `ExperimentData._analysis_results` data format from `ThreadSafeOrderedDict` to the thread safe wrapper of the pandas data frame. This data frame may consists of following columns:
+
+- name (str): Same as `AnalysisResultData.result_type`. This name will appear in the result database.
+- value (Any): Same as `AnalysisResultData.result_data["_value"]`. Any value that QE analysis will generate. This value must be serializable through the `ExperimentEncoder`.
+- chisq (float): Same as `AnalysisResultData.chisq`. Chi-squared value of the entry. Probably this can be moved to extra dict.
+- quality (str): Same as `AnalysisResultData.quality`. Systematically evaluated quality of the fitting to yield this data.
+- device_components (List[DeviceComponent]): Same as `AnalysisResultData.device_components`. Usually `.physical_qubits` of the experiment.
```
chriseclectic, line 67: This field might get unwieldy for experiments on large devices. If we used ints you could get around this by having a `range` object allowed to. For DeviceCompnents we might need to think about making the components support ranges, tuples etc or have some sort of equivalent thing

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+
+In this proposal, we are going to replace the data format of analysis results, based off of our users survey.
+
+### A. Experiment data
+
+We propose to replace `ExperimentData._analysis_results` data format from `ThreadSafeOrderedDict` to the thread safe wrapper of the pandas data frame. This data frame may consists of following columns:
+
+- name (str): Same as `AnalysisResultData.result_type`. This name will appear in the result database.
+- value (Any): Same as `AnalysisResultData.result_data["_value"]`. Any value that QE analysis will generate. This value must be serializable through the `ExperimentEncoder`.
+- chisq (float): Same as `AnalysisResultData.chisq`. Chi-squared value of the entry. Probably this can be moved to extra dict.
```
chriseclectic, line 65: Agree this is probably not necessary as a main column as it is experiment specific.

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+experment_data = pexp.run()
+
+experiment_data.analysis_results(dataframe=True)
+```
+
+Analysis results:
+
+|          | name   | value  | quality | device_components |  backend        | creation_datetime          | experiment | ... |
+|----------|--------|--------|---------|-------------------|-----------------|----------------------------|------------|-----|
+| 03fe91s1 | T1     | 100e-6 | good    | [Q0]              |  ibm_washington | 2022-11-22 03:55:30.313872 | T1         | ... |
```
chriseclectic, line 105: Where do variances/standard errors get shown get shown? As part of the value?

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+
+|          | name   | value  | quality | device_components |  backend        | creation_datetime          | experiment | ... |
+|----------|--------|--------|---------|-------------------|-----------------|----------------------------|------------|-----|
+| 03fe91s1 | T1     | 100e-6 | good    | [Q0]              |  ibm_washington | 2022-11-22 04:01:30.313872 | T1         | ... |
+| 1c4ad2a1 | T1     | 103e-6 | good    | [Q0]              |  ibm_washington | 2022-11-22 05:01:14.123201 | T1         | ... |
+
+
+#### Complement
+
+As these examples show, experimentalist can distinguish the outcomes of batch (with `experiment`) and parallel experiments (with `device_components`). Even though there are multiple entries for the identical parameter (example 3), they have different result IDs and creation date times. In other words, one can sort the entries by `creation_datetime` and get the time-ordered sequence of `value` for the trend analysis:
```
chriseclectic, line 157: These should have different experiment_ids too

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+    x="creation_datetime",
+    y="value",
+)
+```
+
+See pandas documentation for more details.
+
+### B. Introduction of artifacts
+
+Result database allows experimentalists to store `artifact` data per experiment ID. Thus, in QE, we can add `ExperimentData.artifact` to save supplementary data such as raw curve data points. In principle this can be a free-from dictionary, but it would be better to provide the artifact base class and subclasses to support custom JSON serialization to generate the REST API payload and the input data type validation. In the following we describe `CurveAnalysisArtifact` for `CurveAnalysis`. Note that artifact must be the [thread-safe container](https://github.com/Qiskit/qiskit-experiments/blob/b150427f86c74a02d638e06507864c7bb060c27c/qiskit_experiments/database_service/utils.py#L168-L245) subclass. Each analysis class must know the associated artifact subclass to populate the supplementary data.
```
chriseclectic, line 180: Do we definitely want to use JSON for artifacts, or can we support other more efficient file formats too? Eg saving a pandas dataframe directly, or np array, or pickling some things? 

Is there a particular reason we need artifacts to be a thread safe container? These could just be arbitrary things that are generated by an experiement/analysis. eg a plot, a dataframe, a zip file etc. I'm not sure we should make any assumptions about them being thread safe or even safe to modify by other experiments.

Probably we just want a list of supported types for artifacts (like svg, dataframe, array, maybe json/picklable things?)

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+The artifact is a part of `ExpeirmentData`, however, each `BaseAnalysis._run_analysis` method must instantiate and return a list of artifact instances to the experiment data. For example, a composite analysis instance may return multiple artifacts for each qubit or experiment.
+
+
+#### BaseArtifact
+
+The base class of the artifact. Probably Python `dataclass`. This class implements `__json_encode__` and `__json_decode__` methods and subclass can override if necessary.
+This base class may have the field for list of figures. In addition to this, artifact object must provide metadata including
+- experiment (str): Name of experiment that generated this data.
+- device_components (List[DeviceComponent]): List of device components that this data is associated with.
+- creation_datetime (datetime): When this data is created. For example, an experimentalists can sort the artifacts by date to visualize the trend of curves in a heat map in line with time.
```
chriseclectic, line 191: Should artifacts be modifiable, or only able to save, load, delete. If modifiable should probably have a `created` and `modified` datetime fields

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+
+The artifact is a part of `ExpeirmentData`, however, each `BaseAnalysis._run_analysis` method must instantiate and return a list of artifact instances to the experiment data. For example, a composite analysis instance may return multiple artifacts for each qubit or experiment.
+
+
+#### BaseArtifact
+
+The base class of the artifact. Probably Python `dataclass`. This class implements `__json_encode__` and `__json_decode__` methods and subclass can override if necessary.
+This base class may have the field for list of figures. In addition to this, artifact object must provide metadata including
+- experiment (str): Name of experiment that generated this data.
+- device_components (List[DeviceComponent]): List of device components that this data is associated with.
```
chriseclectic, line 190: I don't think this is necessary for a generic artifact. Probably you just need something like `name` `artifact_id` `experiment_id`, `created`, maybe something like `filename/format/file_format` would be useful too for downloading directly from a db for things like images, or csv files, json etc

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+- creation_datetime (datetime): When this data is created. For example, an experimentalists can sort the artifacts by date to visualize the trend of curves in a heat map in line with time.
+
+
+#### CurveAnalysisArtifact
+
+This artifact have two extra fields.
+
+- fit_status: A dictionary (that may provide nice html view) of fitting status. This is currently provided as analysis result entry with the name `@Parameters_*`. This is not really the analysis results and thus moved to artifact. This contains information about fitting, such as fit models, solver status, initial and final fit parameters and so forth. For example,
```
chriseclectic, line 198: Im not confident subclasses is the right way to go for artifacts, it looks more like here you want to define some class that would itself be the value/data of an artifact that is saved, and you want some nice way to html visualize that value.

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+Usually experimentalists are interested in the time at which the experiment was done, rather than a timestamp for the resultDB entry.
+This suggests these time information must be handled separately.
+- fit (dict): A free-form dictionary allowing the storage of arbitrary data.
+This can be used to store the `value`, `extra` and `source` fields.
+
+Two of the proposed fields do not make sense in a resultDB context and won't be saved/loaded but only handled by QE:
+- `created_in_db`: This has been conventionally managed internally in the `ExperimentData`.
+- `experiment`: This may be stored in the `fit` field. Not only experiment UUID, but also human readable representation of experiment is important to help experimentalists with filtering and sorting analysis parameters.
+
+To reduce the amount of work done by QE, we suggest that the experiment service `qiskit-ibm-experiment` will implement
```
chriseclectic, line 235: I would like if the experiment service has its classes and functionality to work with saving/loading, and updating/modifying (where allowed) raw data, analysis results, artifacts (which I think could include figures), and that experiments is basically just initializing and populating the values of these classes to be saved, or modifying loaded ones with as little special handling has possible.

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+indicted using an optional parameter, `dataframe` (bool) which defaults to `False`, inside the methods
+
+- `IBMExperimentService.create_analysis_results()`
+- `IBMExperimentService.update_analysis_results()`
+- `IBMExperimentService.analysis_result()`
+- `IBMExperimentService.analysis_results()`
+
+## Migration plan
+
+Even though this is the change of internal data structure, this may impact community developers who write own experiment on top of our base classes. Thus, the change should be made step by step to guarantee the backward compatibility.
```
chriseclectic, line 260: I think here is where we should try and have fixed classes for the service/db for relevant data types, and the intention is these should not be modified/subclasses, but worked with by other classes within qiskit-experiments or other frameworks.

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+
+`ExperimentData.add_analysis_results()` dispatches the add logic.
+The method signature is updated to `results: Optional[Union[AnalysisResult, List[AnalysisResult]]] = None, **kwargs: Any`.
+If results argument is not None, this must be the legacy analysis result data.
+This implies existing analysis classes can still return this legacy object.
+Legacy result objects are internally converted into a single row of the data frame and added.
+If there are only keyword arguments, we can directly populate the analysis result data frame and bypasses generation of redundant `AnalysisResult` which no longer exists in the `ExperimentData`.
+Existing analysis classes in the Qiskit Experiments namespace will switch to this kwargs style, namely, it must directly populate analysis data from `_run_analysis` method.
+
+`ExperimentData.artifacts()` and `.add_artifacts()` methods are newly implemented with new protected member `ExpeirmentData._artifacts` which is a thread safe list of artifacts.
```
chriseclectic, line 275: If we define an artifact class in the service, could we migrate figure handling to use artifacts? Maybe this requires too much change of the web UI to do in one go

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+- results argument from the `ExpeirmentData.add_analysis_results()`.
+- dataframe argument from the `ExpeirmentData.analysis_results()`.
+
+
+## Alternative Approaches
+N/A
+
+## Questions
+- What data frame columns must be shown to end users.
+- How we expose `extra` column? A single column of a free-form dict is enough? We cannot apply convenient data frame filtering methods to such column. This must be flattened? This may cause problem when we have multiple experiments, i.e. BatchExperiment. Every experiment may have different extra dict keys.
```
chriseclectic, line 316: Maybe we could have defined columns for the service, and then a compacted column (ie extra), that can be expanded / collapsed when loading and saving? For batch experiments this might involve a lot of columns as the union of all individual results, where some would need to have a empty/null/ N/A field to distinguish which should be discarded when collapsing back down.

If we do something like this im not sure if this functionality should be in the service, or qiskit experiments. (It could probably be done with custom accessors too)

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,320 @@
...
+- `AnalysisResult` and `AnalysisResultData`.
+- results argument from the `ExpeirmentData.add_analysis_results()`.
+- dataframe argument from the `ExpeirmentData.analysis_results()`.
+
+
+## Alternative Approaches
+N/A
+
+## Questions
+- What data frame columns must be shown to end users.
```
chriseclectic, line 315: Maybe we can use custom pandas accessors either in service or qiskit experiments for deciding which columns to show.

0000-experiment-dataframe.md
nkanazawa1989: What is the purpose of `modified` field? This is a metadata to control autosave?

nkanazawa1989: `created` sounds like a boolean value. I renamed `device_components` -> `components`, and `creation_datetime` -> `date`. Note that creation date time value (date time object repr) is rather lengthy, and this change doesn't shorten the widths of the cells.

0000-experiment-dataframe.md
nkanazawa1989: This is bit problematic because some experiments target non qubit components, such as resonators. I agree we must update `DeviceComponents` so that it can take slice value. But I'm not sure if we will have such experiment with large qubit set. In the parallel experiment, the device components here show qubits of each nested experiment, so qubit size is typically one or two.

0000-experiment-dataframe.md
nkanazawa1989: Done in [8454c5e](https://github.com/Qiskit/rfcs/pull/28/commits/8454c5eeb0a4ad360ef7966b56c5e67e1471b369)

0000-experiment-dataframe.md
nkanazawa1989: Yes. This not a serialized version, and raw UFloat object will come here. So variance will be shown through its repr.

0000-experiment-dataframe.md
nkanazawa1989: Thanks. Updated in [8454c5e](https://github.com/Qiskit/rfcs/pull/28/commits/8454c5eeb0a4ad360ef7966b56c5e67e1471b369)

0000-experiment-dataframe.md
nkanazawa1989: I don't have any use case that we need to modify the artifact. If an experimentalist needs different visualization, probably that will be new artifact entry.

0000-experiment-dataframe.md
nkanazawa1989: ^ @gadial

0000-experiment-dataframe.md
nkanazawa1989: ^ @gadial

0000-experiment-dataframe.md
nkanazawa1989: > For batch experiments this might involve a lot of columns as the union of all individual results, where some would need to have a empty/null/ N/A field to distinguish which should be discarded when collapsing back down.

Yes, that's my concern. Probably, QE can always format data frame so that it only returns union of columns to end users, e.g.

```
entry1: ["value", "units", "chisq", "components", ...]
entry2: ["value", "units", "chisq", "components", ...]
entry3: ["value", "components", ...]
```

When a user request data of entry1 and entry2, QE (`ExperimentData`) returns `"value", "units", "chisq", "components",...`.  When one requests 1 and 3, only `"value", "components", ...` are returned.

0000-experiment-dataframe.md
nkanazawa1989: Probably we have assume different data structure for artifact. I assumed following:
```
Artifact: List[BaseArtifact]
```
and 
```
BaseArtifact: dataclass
```
where single analysis instance returns a single formatted artifact object. So data type validation is done by each artifact instance and artifact can contain multiple data (e.g. figure, curve data, fit status, ...).

Likely you are assuming following:
```
Artifact: List[Any]
```
where a single analysis instance generates multiple data entries, and it can push them to the artifact without structuring the data. Then each entry is managed by the experiment id and name. For example, parallel experiment of the same kind of experiment can push merged curve data as a single data entry. The artifact object does validation for acceptable data type. I like your approach.

0000-experiment-dataframe.md
nkanazawa1989: This must be resolved by redesigning the data structure. We can create artifact entry per data rather than per analysis. https://github.com/Qiskit/rfcs/pull/28/commits/263e94a3c5a02e3be14c7916652590b63e970ba7

0000-experiment-dataframe.md
nkanazawa1989: I redesigned the artifact. Seems like new approach is better than that I originally proposed here. https://github.com/Qiskit/rfcs/pull/28/commits/263e94a3c5a02e3be14c7916652590b63e970ba7

0000-experiment-dataframe.md
gadial: Currently `qiskit-ibm-experiment` uses the `ExperimentData` and `AnalysisResultData` dataclasses, which are also used inside `qiskit-experiments`. I have no objection to modifying how these dataclasses look or used, but this is really a question of what `qiskit-experiments` users would use.

We can also drop these classes altogether and stick with padas dataframes (that's basically what the local DB is doing now). Again, the question is what users would prefer.

0000-experiment-dataframe.md
gadial: The service already know how to handle figures, and should definitely be able to handle artifacts as well. I agree that delegating as much possible code to the service is preferable.

0000-experiment-dataframe.md
gadial: I agree, the question is how the minimal interface `qiskit-experiments` uses would look like; I can make the changes in the `qiskit-ibm-experiment` to accommodate for (hopefully) anything we'd like.

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,404 @@
...
+
+|          | name   | value  | quality | components |  backend        | run_time                   | experiment | ... |
+|----------|--------|--------|---------|------------|-----------------|----------------------------|------------|-----|
+| 03fe91s1 | T1     | 100e-6 | good    | [Q0]       |  ibm_washington | 2022-11-22 04:01:30.313872 | T1         | ... |
+| 1c4ad2a1 | T1     | 103e-6 | good    | [Q0]       |  ibm_washington | 2022-11-22 05:01:14.123201 | T1         | ... |
+
+
+#### Complement
+
+As these examples show, experimentalist can distinguish the outcomes of batch (with `experiment`) and parallel experiments (with `components`). Even though there are multiple entries for the identical parameter (example 3), they have different result and experiment IDs and run times. In other words, one can sort the entries by `run_time` and get the time-ordered sequence of `value` for the trend analysis:
```
coruscating, line 151: I'm thinking about cases of complicated nested composite experiments where looking at `experiment` and `components` are not enough to distinguish between different results. The user might be interested in is something like "I want the result of the fourth T1 experiment that was run simultaneously on Q0 and Q10." Of course, this can be deduced by looking at all the columns, but are we expecting the users to be familiar with pandas and write their own multi-column filters?

nkanazawa1989: Good point. Such functionality can be provided through the pandas accessor, but the good point of using pandas is they can find many well written documentation in the web. As you know we don't have enough capability of maintaining detailed documentation and I'd like to avoid overengineering at this stage (even if we provide the logic it is tough for users to actually use it). Can you provide some configuration that might mess up the result table?

coruscating: Yeah, I don't think we should make more work for ourselves implementing too many custom accessors since pandas already has many features and is well-documented. Perhaps a tutorial showing some particularly useful operations would be good enough. As we discussed, the main issue I can think of is `run_time` can be potentially out of order from the actual execution time on the device. I can think of a few times that are relevant:
1. time when job was submitted to the queue
2. time when job actually started to run
3. time when job finished running
4. `run_time` as defined here, time of experiment data receiving the job results

Usually 1 and 2 will be in the same order, but experimentalists who have access to the queue can potentially move jobs around in it. So for something like "get the result of the fourth T1 experiment", I wonder which time metric we should be using, or should we be recording multiple times and ask the user to specify what they want.

chriseclectic: Maybe it is worth thinking a bit more about how we should record times.

For experiments I would expect the important times to be 2 and 3, when it started running on the device, and when it finished running on the device. However these times are generally specific to backend data (count/iq), not individual analysis results so any results processed from the same batch of data, such as parallel experiments, will have the same times. I don't think the specific time that ExperimentData received a job result is particular relevant (unless the job contains no timing information for its execution).

Another idea would be to include a `created_time` for analysis results like you have for artifacts. I'm imagining this could be relevant if you re-ran analysis on exisiting data (which would have a fixed time for when it ran) and wanted to save both results to the DB (im not sure this is allowed through qiskit-experiments without the two results also having different experiment ids, but you could probably do it directly through service).

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,444 @@
...
+  name="my_second_plot",
+  data=my_plot2,
+)
+```
+
+In this case, the user can bypass creation of `AnalysisResult` or `ArtifactData` since these are assumed to be a container used by an experiment author to write `_run_analysis`.
+
+### C. Updating existing results or artifacts.
+
+An end user (experimentalist) can modify analysis result or artifact after experiment.
```
chriseclectic, line 328: Is this intended to replace the artifact with a new one using the same `artifact_id` or to delete the existing one and add the replacement artifact with a new `artifact_id`?

0000-experiment-dataframe.md
```diff
@@ -0,0 +1,444 @@
...
+```python
+exp_data = MyExperiment(...).run().block_for_result()
+save_status = exp_data.save(blocking=False)
+if save_status.running():
+    save_status.block()
+for error in save_status.errors():
+    print(error)
+```
+
+Note that the experiment service already supports API to save artifacts. The payload must be JSON format.
```
chriseclectic, line 375: Should this JSON serialization happen in the service layer, so the python artifict object has a python class as its value? If the later who's responsibility is the JSON serializer in this case? qiskit experiments or the service?

0000-experiment-dataframe.md
nkanazawa1989: I think it depends on the experiment service implementation, but I was thinking to replace existing entry with the same ID.

0000-experiment-dataframe.md
nkanazawa1989: Good point. I think this is the responsibility of service since user may want to implement custom (local) service that is capable of handling custom data format. Indeed the conventional IBM provider implements [own encoder](https://github.com/Qiskit/qiskit-ibmq-provider/blob/master/qiskit/providers/ibmq/utils/json_encoder.py) rather than in Qiskit terra. I think we should move current experiment encoder to service.

nkanazawa1989: I clarified this in [3de30ba](https://github.com/Qiskit/rfcs/pull/28/commits/3de30ba6235fb53d42881029f2b729cc7dc380e4).

0000-experiment-dataframe.md
nkanazawa1989: Thanks Chris. I noticed that Job object reports multiple timestep, i.e.
```python
job = backend.run(qc)
job._wait_for_completion()
job.time_per_step()
```
this returns a dictionary
```
{'CREATING': datetime.datetime(2023, 1, 26, 15, 22, 4, 465000, tzinfo=tzlocal()),
 'CREATED': datetime.datetime(2023, 1, 26, 15, 22, 5, 813000, tzinfo=tzlocal()),
 'VALIDATING': datetime.datetime(2023, 1, 26, 15, 22, 6, 43000, tzinfo=tzlocal()),
 'VALIDATED': datetime.datetime(2023, 1, 26, 15, 22, 6, 629000, tzinfo=tzlocal()),
 'QUEUED': datetime.datetime(2023, 1, 26, 15, 22, 6, 728000, tzinfo=tzlocal()),
 'RUNNING': datetime.datetime(2023, 1, 26, 15, 22, 7, 752000, tzinfo=tzlocal()),
 'COMPLETED': datetime.datetime(2023, 1, 26, 15, 22, 19, 271000, tzinfo=tzlocal())}
```
I think we can pick RUNNING time as a `run_time`, and separately add `created_time` to cover your case of reanalysis. Updated rfc with [9a62098](https://github.com/Qiskit/rfcs/pull/28/commits/9a62098d63278310863c9d1c34c431f4fa19cee7).

chriseclectic (approved):
For me this looks good to start moving to the next stage of prototyping

nkanazawa1989:
Thanks @t-imamichi for the suggestion. We are aware of that package but we intentionally chose pandas for now. The number of analysis result entries in `ExperimentData` is estimated to be roughly ~100 and performance gain with Rust implementation is not quite drastic, and we want to chose a popular package with lot of documentation so that we can avoid maintenance/support overhead.

PR #29, opened 2022-10-26
A proposal to improve our documentation workflow

javabster:
Our current documentation workflow is not the most efficient and gives rise to many inconsistencies across our many Qiskit repos. This rfc aims to start conversation about how we could improve this process

### DIFF
--- a/####-rfc-docs
+++ b/####-rfc-docs
@@ -1,0 +1,126 @@
+# RFC Title
+
+| **Status**        | **Proposed/Accepted/Deprecated** |
+|:------------------|:---------------------------------------------|
+| **RFC #**         | ####                                         |
+| **Authors**       | Abby Mitchell (abby.s.mitchell@ibm.com)      |
+| **Submitted**     | 2022-10-26                                   |
+| **Updated**       | YYYY-MM-DD                                   |
+
+## Summary
+This proposal is for a restructure of our current documentation workflow. This document outlines the current procedure, highlights its limitations, and proposes an alternative for discussion.
+
+In short, the suggestion is to replace our existing workflow with this:
+
+- Individual Qiskit repos start running the sphinx JSON builder instead of the html builder (so the output is `json` files instead of `html`)
+- Individual Qiskit repos deploy their built `json` files somewhere that can be accessed by the `qiskit.org` repo
+- The qiskit.org repo integrates the documentation pages into it's development and deployment pipeline
+- The `qiskit_sphinx_theme` is sunset and Qiskit teams don't need to handle any of the frontend aspects of their docs anymore
+
+
+## Context - the current docs status quo
+
+The existing workflow for documentation (for each Qiskit repo) broadly follows this process:
+
+- Individual devs write docstrings in the files they touch during their regular work
+- At release time, as part of the deployment github workflow (usually triggered from `.github/workflows/docs-deployment`), a [sphinx](https://www.sphinx-doc.org/en/master/) build is run which generates a `_build/html` folder. The html files generated are then deployed to COS.
+- After this each individual html page is available for users to view at qiskit.org/documentation/path-to-file.
+- Each repo is responsible for maintaining their own sphinx build and deployment scripts
+- The UI styling is set by the `qiskit_sphinx_theme`, a pip installable package that each repo should be using during the sphinx build to enforce the styles. However, in reality many repos overwrite this theme in their `docs/_templates` folder. Additionally the `qiskit_sphinx_theme` repo has not been actively maintained until recently (the Qiskit Developer Adovacy squad is looking at refactoring this).
+
+### What are the drawbacks of this process?
+
+- Lack of consistency in UI/UX between qiskit.org and qiskit.org/documentation. The Qiskit Digital team implements the frontend for qiskit.org, qiskit.org/learn etc. using the Vue framework and a standard library of IBM Carbon components whereas qiskit.org/documentation mainly uses plain html and css (which is limited when it comes to more complex features, and requires more manual effort to style consistently).
+- Lack of consistency in UI/UX between different Qiskit repos due to each team overriding the `qiskit_sphinx_theme` slightly differently.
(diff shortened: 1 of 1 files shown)


####-rfc-docs
```diff
@@ -0,0 +1,126 @@
...
+
+<img width="1785" alt="Screenshot 2022-10-25 at 5 58 03 PM" src="https://user-images.githubusercontent.com/23662430/198141449-4a2298a8-9796-4105-a25b-9fa20ece4a0d.png">
+
+### Challenges/Limitations so far
+
+some initial challenges/drawbacks have been identified as part of the investigation phase so far:
+- each `json` file has a `body` attribute which contains the main content of the page (i.e. what gets rendered in the middle section of the screenshot). This attribute is a long html string that we can render with Vue BUT we need to go manually change all the styling. Doing this for every possible html tag and class name used in all the docstrings will be a large piece of work. The screenshot is pretty similar to the raw output, I haven't adjusted much of the css there.
+- Internal links within the `body` object get added to the html like this: `href="../../stubs/qiskit.circuit.QuantumCircuit/#qiskit.circuit.QuantumCircuit" `, which is not ideal, we’ll have to figure out how to adjust that for every link on every page
+- The sphinx build output is actually `.fjson` which is exactly like `.json` but with an f in front of it. This could maybhe cause issues?
```
mtreinish, line 99: FWIW,The built in json builder (along with the other built-in serialized builder using pickle) is basically designed to host the content from a custom web app instead of being consumed as static content on a web server directly. That's why it has the html body embedded as an object field. You might be better off just writing a custom build to output content the way you want if this is the route you're pursuing.

####-rfc-docs
```diff
@@ -0,0 +1,126 @@
...
+- Performance: the sphinx build produces 1 json file per documentation page - thats a LOT of pages that qiskit.oprg will need to render and deploy, which could cause performance issues on qiskit.org
```
mtreinish, line 100: This is true of the html output too, it's why the main docs build upload/sync takes so long. There is a single page view but given the amount of content that ends up being gigantic too.

mtreinish (commented):
TBH, this proposal seems way too big to me. You're proposing building a custom content builder system on top of sphinx (which is kind of it's job). I feel like you'll end up with a lot of extra effort and a long term maintenance nightmare if you go down such a custom route. As we'll be tied to a bespoke pipeline to host anything. My other big concern with doing something like this is that for the people writing the content in the qiskit projects as part of the development. If you divorce the actual static content building from the sphinx build itself you basically make it impossible for people writing new documentation to know exactly how things will look without the full web dev pipeline. Especially if the sphinx output in a normal docs build ends up looking very dfifferent from what gets hosted. I would personally find this very frustrating, especially if formatting bugs were reported in the docs.

What about something a bit smaller in scope, like just writing a custom sphinx html theme package from scratch instead of using the inherited mess of the current qiskit-sphinx-theme <- pytorch-sphinx-theme <- sphinx-rtd-theme. We're not tied in any way to that package, and if it's easier to just abandon it and start from scratch with more modern tooling we should go for it. I think you should be able to do this without too much trouble as it's all static content at the end of the day and sphinx just treats the theme as a template for generating those file. That being said the sphinx documentation around html theming is quite lacking (TBH, like all of sphinx, you'd think the documentation package would have better documentation): https://www.sphinx-doc.org/en/master/development/theming.html .

I did some quick searching and found someone who made a custom sphinx theme package with vuepress and vue.js, a demo is published here: https://romnn.github.io/sphinx_press_theme/ and the code is here: https://github.com/romnn/sphinx_press_theme
I'm not up to date on all the web frontend frameworks that are popular nowadays, and I know there are subtle differences between things I often gloss over, so this might not be a relevant example.

1ucian0:
>  If you divorce the actual static content building from the sphinx build itself you basically make it impossible for people writing new documentation to know exactly how things will look without the full web dev pipeline.

The developer writing documentation should focus on the content of `article` (the main content body) and not the toc, menus, footers, and the like. I agree than provider a way for developers to render that as it will be published is necessary. 

The current approach of static pages makes it impossible to integrate different built instances. Take the example of versioning. Once you clicked in a previous version (say you are in 0.39 and you clicked in 0.38) there is not way back (in this case, to 0.39)

https://user-images.githubusercontent.com/766693/199343589-3ee2ffe8-786c-4310-90ac-f9d23c48b884.mov

With the current approach, there is no solution to this problem, because once pages are created, they cannot be changed. The PyTorch approach was  https://pytorch.org/torchx/versions.html, but even there it is not consistent along components (https://pytorch.org/torchrec/ has no version selector, or shows the version at all). Similarly, also in PyTorch there is no consistency on the top menu. 

https://user-images.githubusercontent.com/766693/199345874-dee4763f-8872-41ec-b3b5-44e7c38efbaf.mov

Hopefully this shows that we need a way to disentangle content from layout.

mtreinish:
You're conflating a few different things which aren't really related to the point I was making before. Static content hosting is a requirement for how we host the website and there is no option for an interactive dynamic web application to run code server side. This isn't going to change and is also not what this proposal is about (at least as I read it).

The limitations you're pointing at aren't inherent to static content, you can build a solution around having shared version links fairly easily using a reference to a shared file in a consistent relative path very easily. The specific issue about versions tables not being updated for past releases that you're pointing out there is because we evaluate the versions at build time, which was a specific decision in Qiskit/qiskit#1261. We discussed how to do this when we added the stable version documentation originally and I laid out a rough sketch of how to do it (you can see the start of work on this earlier in Qiskit/qiskit#574 but we never finished it as it wasn't a priority), but nobody had the bandwidth to implement it at the time and we settled on doing the faster to implement solution which got us most of what we wanted.

The problem with saying the content writer/developer shouldn't have to worry about the layout is that they're actually far more coupled than it might first appear given how sphinx works. The exact method the docstrings get rendered together is by filling out templates with the content from the actual documentation and what the exact layout looks like. This doesn't actually change in this proposal either it just changes the mechanism by which this happens. For example the contents of the document determine what gets populated in table of contents which end up in side bars, etc. Similarly theme decisions can impact how things get rendered and can inform decisions around how you choose to format elements in the documentation. To do a good job writing documentation you need to be able to build the final compiled output locally, and adding an entire parallel JS build framework for doing that is pretty much a non-starter from my PoV; people already have enough issues with building docs to add the complexity on top of it.

Eric-Arellano:
We've decided to not pursue this Chimera approach for now. We are sticking to using only Sphinx, and migrating to Furo so that the code is more maintainable and we have better front end like being more responsive on mobile.

@javabster can you please close this?

PR #30, opened 2023-03-15
Preliminary representation of rvalue classical expressions in Terra

jakelishman:
This is a proposal for initial representation of classical expressions in Terra.  It is simple bit-level operations only on effectively `Clbit` and `ClassicalRegister`, and is deliberately designed to be separable from the rest of Terra so it can be replaced in the future if necessary, without painful decoupling of these objects from core Terra ones.

This proposal notably does not require any new public methods or trackers on `QuantumCircuit`, `Clbit` or `ClassicalRegister`.  This includes _not_ having a way to store a runtime expression in a variable of `QuantumCircuit` for re-use, nor of using these expressions in the parameters to gates.  These may come in the future, but are deliberately excluded from prelim exploratory representations.

### DIFF
--- a/0030-simple-classical-representations.md
+++ b/0030-simple-classical-representations.md
@@ -54,8 +54,8 @@ The allowed expressions will be the following:
   - runtime `bool` (`Clbit`);
   - runtime `uint[n]` (`ClassicalRegister(size=n)`).
 - Unary:
-  - `uint[n] -> uint[n]`: `~` (bitwise negation);
-  - `bool -> bool`: `!` (boolean negation).
+  - `uint[n] -> uint[n]`: bitwise negation
+  - `bool -> bool`: Boolean negation
 - Binary:
   - `uint[n] * uint[n] -> uint[n]`: `&`, `|`, `^`;
   - `bool * bool -> bool`: `&&`, `||`;
--- a/0030-simple-classical-representations.md
+++ b/0030-simple-classical-representations.md
@@ -1,8 +1,8 @@
 # Preliminary representation of rvalue classical expression in Qiskit
 
-| **Status**        | **Proposed/Accepted/Deprecated** |
+| **Status**        | **Accepted** |
 |:------------------|:---------------------------------------------|
-| **RFC #**         | ####                                         |
+| **RFC #**         |  30                                        |
 | **Authors**       | Jake Lishman (jake.lishman@ibm.com)          |
 | **Deprecates**    | None                                         |
 | **Submitted**     | 2023-03-15                                   |
@@ -308,10 +308,6 @@ This function-decorator method is not chosen because:
 
 In the context of a larger circuit-builder interface, where the entirety of the `QuantumCircuit` is built up using such functions, many of the negatives of this form are reduced.
 
-
-## Questions
-
-?
 
 ## Future Extensions
 
--- a/0030-simple-classical-representations.md
+++ b/0030-simple-classical-representations.md
@@ -1,4 +1,4 @@
-# RFC Title
+# Preliminary representation of rvalue classical expression in Qiskit
 
 | **Status**        | **Proposed/Accepted/Deprecated** |
 |:------------------|:---------------------------------------------|
--- a/0030-simple-classical-representations.md
+++ b/0030-simple-classical-representations.md
@@ -120,6 +120,24 @@ bounded_reg = expr.And(
 These expressions will then be given directly to `IfElseOp` or `WhileLoopOp` in their `condition` fields.
 These expressions will not be valid in the general `Instruction.condition` field; we do not want to expand any support for something we are already moving to remove.
 
+For example, a complete condition construction might look like (eliding `expr.Value` calls that could be inferred):
+```python
+from qiskit.circuit import QuantumCircuit, ClassicalRegister, QuantumRegister
+from qiskit.circuit.classical import expr
+
+cregs = [ClassicalRegister(3), ClassicalRegister(3)]
+qc = QuantumCircuit(QuantumRegister(3), *cregs)
+
+qc.h(0)
+qc.cx(0, 1)
+qc.cx(1, 2)
+qc.measure([0, 1, 2], cregs[0])
+qc.measure([0, 1, 2], cregs[1])
+with qc.if_test(expr.And(expr.Less(0, cregs[0]), expr.LessEqual(cregs[0], cregs[1]))):
+    # This is the same as the `bounded_reg` comparison above.
+    pass
+```
+
 
 #### Stage 2
 
@@ -138,7 +156,7 @@ with expr.build():
 
 It is not possible to overload the behaviour of the Python Boolean operators (`not`, `and` and `or`) for classes, which is why the bitwise operations `~` and `&` are used instead (with the corresponding precedence frustrations).
 Within the `expr.build()` context, the magic methods of `ClassicalRegister` and `Clbit` will be temporarily overridden to be an eager builder interface for the syntax tree given above.
-This means that the expressions will not need to be bound to variables in order to use them, as the resulting objects will be well-defined and the correct tree form.
+This means that `expr.build()` context can just contain the entire circuit construction without any changes to `QuantumCircuit`; nothing magic happens on exit of the builder context except for tidying up the monkey-patched methods.
 For example, it will be valid to do
 
 ```python
--- a/0030-simple-classical-representations.md
+++ b/0030-simple-classical-representations.md
@@ -6,7 +6,7 @@
 | **Authors**       | Jake Lishman (jake.lishman@ibm.com)          |
 | **Deprecates**    | None                                         |
 | **Submitted**     | 2023-03-15                                   |
-| **Updated**       | 2023-03-15                                   |
+| **Updated**       | 2023-03-22                                   |
 
 
 ## Summary
@@ -223,6 +223,26 @@ Pull request 5 can be permitted to follow in a subsequent Terra release, if nece
 
 ## Alternative Approaches
 
+### Reuse of `ParameterExpression`
+
+We have talked about using the existing `ParameterExpression` before, for this.
+This poses several problems, because overloading it from its current use as a _compile-time_ stand-in for unknown values to involve arbitrary _run-time_ expressions opens the door for many errors where those two domains don't fully align.
+`ParameterExpression` is also build on `sympy`/`symengine`, which don't have any sort of rigorous type system for us to use.
+We would have to hack one on top of them, which would make a lot more of the manipulation interfaces around these types far more complex to use.
+These also perform (by default) aggressive simplifications based on arithmetic operations on the reals; this does not translate cleanly to storing operations on mixed types, or having casts from one type to another, and so on.
+
+### Reuse of `classicalfunction.BooleanExpression`
+
+This type is somewhat more similar to the immediate goals of this PR, but it:
+
+- has hard requirements on being a Python AST walker, which hurts ergonomics for construction of simple conditions (see next comment, and the "call" syntax to apply to given objects would need to implemented still);
+- relies on `tweedledum`, which is unmaintained (which is why `qiskit.circuit.classicalfunction` is pending complete removal from Terra);
+- cannot easily allow dynamic construction of conditions, since it relies on being a static analyser.
+
+It is also currently only defined for bit types, not the integer types we need for `ClassicalRegister` (despite the existence of `Int2`).
+This possibly could be extended, but the other points above are enough reason to dismiss this.
+
+
 ### AST-visitor builder interface
 
 The stage-2 builder interface for the expression syntax given above had a problem that Python does not allow classes to override the behaviour of binary Boolean operations.
@@ -245,9 +265,11 @@ This is possible with function decorators because the decorator can access the P
 
 This function-decorator method is not chosen because:
 
-- in practice, since it only applies to expressions and is not part of a larger circuit builder interface, it ends up being much more verbose due to the need to first define, then call
+- in practice, since it only applies to expressions and is not part of a larger circuit builder interface, it ends up being much more verbose due to the need to first define, then call;
 - it is harder to write and maintain a Python AST visitor, which makes it less ideal for a preliminary implementation
-- it has to do type checking twice; once during the AST visit, and once during the subsequent object call
+- it has to do type checking twice; once during the AST visit, and once during the subsequent object call.
+
+In the context of a larger circuit-builder interface, where the entirety of the `QuantumCircuit` is built up using such functions, many of the negatives of this form are reduced.
 
 
 ## Questions
--- a/0030-simple-classical-representations.md
+++ b/0030-simple-classical-representations.md
@@ -57,7 +57,7 @@ The allowed expressions will be the following:
   - `uint[n] -> uint[n]`: bitwise negation
   - `bool -> bool`: Boolean negation
 - Binary:
-  - `uint[n] * uint[n] -> uint[n]`: `&`, `|`, `^`;
+  - `uint[n] * uint[n] -> uint[n]` and `bool * bool -> bool`: `&`, `|`, `^`;
   - `bool * bool -> bool`: `&&`, `||`;
   - `uint[n] * uint[n] -> bool`: `==`, `!=`, `<`, `<=`, `>`, `>=`.
 
@@ -315,5 +315,5 @@ In the context of a larger circuit-builder interface, where the entirety of the 
 
 ## Future Extensions
 
-- An entire classical system including the ability to assign runtime expressions to variables that are managed by `QuantumCircuit`.
+- An entire classical system including casting between different types, and the ability to assign runtime expressions to variables that are managed by `QuantumCircuit`.
 - An extension of the type system to include values that can be used as the operands of gate objects.
--- a/0030-simple-classical-representations.md
+++ b/0030-simple-classical-representations.md
@@ -19,12 +19,12 @@ This is not required to be the final design for classical handling in Terra, it 
 
 - Current hardware is gaining support for runtime operations and subsequent conditions on classical bits
 - IBM dynamic-circuits hardware can do bit-level calculations, but there is no way to represent this in Qiskit to pass it on
-- On Terra, we need more of a feel for what will actually help us represent more complex classical constructs
+- In Terra, we need more of a feel for what will actually help us represent more complex classical constructs
 
 
 ## User Benefit
 
-- Users of IBM dynamic circuits will be able to take advantage of the classical-processing capabilities of those compilers from within Qiskit
+- Users of IBM dynamic circuits will be able to take advantage of the classical-processing capabilities of those systems from within Qiskit
 - This support is not limited to IBM so may extend further, but we are not aware of other dynamic-circuits hardware accessible through Qiskit
 - For Terra development: this is a "build one to throw away"-type sytem, so we can see in actual usage what works and doesn't for us before we commit to further classical processing.
 
@@ -82,7 +82,7 @@ All of these points are eligible to be included in future work, just not the MVP
 
 #### Stage 1
 
-This is a temporary representation, so it is a non-goal to produce the best possible UX for constructing these objects.
+This is an experimental representation, so it is a non-goal to produce the best possible UX for constructing these objects.
 In the initial representation phase, users will be required to build the expression-tree representations themselves.
 This will look _something_ like (names up for discussion):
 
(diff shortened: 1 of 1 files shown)


jakelishman:
I just realised that I didn't include QPY support in the RFC, but I think for the MVP use-case we can actually get away without it; the known places that would support these dynamic-circuits paths (i.e. just IBM's `qiskit-ibm-provider` atm) export to OQ3 on the client side.

mtreinish:
> I just realised that I didn't include QPY support in the RFC, but I think for the MVP use-case we can actually get away without it; the known places that would support these dynamic-circuits paths (i.e. just IBM's `qiskit-ibm-provider` atm) export to OQ3 on the client side.

Unless something has changed recently it actually does the OQ3 export server-side and submits the circuit as a qpy payload to the ibm runtime environment which calls the op3 export as part of the runner program.

jakelishman:
Oh no sorry, you're right and I'm totally misremembering.  I'll update to add QPY support as a requirement for release.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+        qc.h(0)
+```
+
+The reason to require entering the builder context to make the magic methods available is for potential future ease of change/removal/deprecation.
+The mutation of global state required to do this makes the builder interface non-thread-safe, but this is not judged to be a problem; it is uncommon to attempt to multithread during manual circuit construction, and the advantages of easy removal/representation change completely outweigh the potential concern.
+
+
+## Detailed Design
+
+The new functionality will live in in `qiskit.circuit.classical`.
```
kevinhartman, line 150:
```suggestion
The new functionality will live in `qiskit.circuit.classical`.
```

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+# file: qiskit/circuit/classical/expr.py
+
+class Expression(ABC):
+    resources: typing.Set[Clbit]
+    type: qiskit.circuit.classical.Type
+```
+
+The `resources` field is effectively a `use-def` tracker, which will more easily enable the conversion to data-flow (`DAGCircuit`) format to evaluate the necessary wires of an instruction.
+The `type` field is a resolved type of the contained expression.
+In this initial release, the only time that `type` will not be a fully qualified type will be when representing a `expr.Value` of a `uint` literal.
```
kevinhartman, line 166: Not fully-qualified in that the size isn't fixed?

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+
+PR 1:
+- write representation of `Expression` and `Type`
+- test that these objects can be constructed and handle type checking correctly
+
+PR 2 (depends on 1):
+- add support to `Expression` of evaluated type `Bool` to `IfElseOp.condition` and `WhileLoopOp.condition`
+
+PR 3 (depends on 2):
+- add support for exporting these `Expression` conditions to the OpenQASM 3
```
kevinhartman, line 211:
```suggestion
- add support for exporting these `Expression` conditions to OpenQASM 3 
```

kevinhartman (commented):
Looks very cool, overall.

Despite being verbose, I'm a fan of the explicit construction of expressions in Stage 1.  Personally, I'm not fully sold on the use of operator overloading in Stage 2. I tend to find it somewhat confusing as a user, since it feels like a surprise when types get magically coerced.

I'm sure you've already considered this, but just in case you haven't, I wonder what you'd think about an expression building syntax that uses chaining.

Here's your example modified to demonstrate what this might look like. Note that this would be in place of using context builders:
```py
from qiskit.circuit import Clbit, ClassicalRegister
from qiskit.circuit.classical import expr

loose = [Clbit() for _ in [None]*3]
cr1 = ClassicalRegister(2)
cr2 = ClassicalRegister(2)

# Equivalent to the current condition `(cr, 2)`:
equal_reg = expr.Value(cr1).eq(2)

# Equivalent to the current condition `(loose[0], False)`
equal_bit = expr.Value(loose[0]).eq(False)

# The above example, but using a Bool-valued unary expression:
not_bit = expr.Value(loose[0]).not()

# The comparison 0 < cr1 <= cr2
bounded_reg = expr.Value(0).lt(cr1).and(expr.Value(cr1).lt_eq(cr2))
```

Arguments passed to chaining functions could be either of type `Expr`, or they could be our standard types, which would be coerced into `Expr`. This is similar to what you propose with operator overloading, but since this only happens for chaining function args, it might give the user better context clues to what is going on (and save an indentation!). Perhaps you could also take advantage of the left-hand side's greater precedence to coerce the right-hand side to match the width in the case of fixed-width type comparisons, though I haven't thought this through.

Assuming you've already considered something like this and have a preference for the proposal as it is, I'm on board with this moving forward as is.

jakelishman:
I hadn't actually considered chaining at all - honestly, I hadn't really considered anything much beyond what I wrote because I was rushing to get it up before the meeting yesterday.  My rough thoughts on it are:

1. it's definitely acceptable per the goal of "we can tear this out again";
2. it doesn't have the problem that the magic methods have of not being able to override the Python logical operators;
3. I'm not certain it's significantly less verbose than just using all the methods directly, which was the main UX-y goal of the builder interface + overload;
4. I'm not fully sold that expression trees are best represented as chains. I feel like they have a habit of over-promoting the left operand of a binary expression - `a.and(b)` _looks_ asymmetric, and it's worse when `a` and `b` are more complex;
5. (minor) there's a few method names in the chain that wouldn't be allowed - `and`, `not` and `or` would need different names, because they're Python keywords.

On balance, I think I personally still prefer the magic-method arithmetic form, but I'm happy to go with the crowd.

0030-simple-classical-representations.md
jakelishman: yeah, "qualified" doesn't seem like the right word here.  I'm pretty sure there is a correct word for it, I just can't think of it.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+For typing in this initial draft, we will make an exact bijection:
+
+- `ClassicalRegister(size=n)` <=> `uint[n]`
+- `Clbit` <=> `bool`
+
+Python Booleans (`True` and `False`) will be used as the `bool` literal values, as they currently are in `Instruction.condition`.
+Non-negative Python ints will be used as the `uint` literal values.
+These do not have an inherent width.
+The width of a literal `uint[_]` will be inferred and fixed to make a containing binary expression validly typed, and mixed-width operations will be forbidden.
+A binary expression in this representation _must_ contain at least one type of fixed width; it will be an error to attempt to use the runtime logic on two literals (this simplifies the implementation, and we won't do constant folding).
```
kdk, line 66: Is there something we can say here about how the width will be inferred? Is this behavior that's coming from OpenQASM3?

kdk: I'm not sure how to interpret "it will be an error to attempt to use the runtime logic on two literals" in this context. Can you explain a little more?

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+Points 1 through 3 are rejected in this initial proposal for the sake of keeping the number of expression-tree nodes and type-system entries small.
+Point 4 is rejected because it would require the definition of additional `Operation` objects and potentially modifications to the `QuantumCircuit` class that we are not yet prepared to make.
+All of these points are eligible to be included in future work, just not the MVP.
+
+
+### Construction
+
+#### Stage 1
+
+This is a temporary representation, so it is a non-goal to produce the best possible UX for constructing these objects.
```
kdk, line 85:
```suggestion
This is an experimental representation, so it is a non-goal to produce the best possible UX for constructing these objects.
```
Maybe?

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+not_bit = expr.Not(expr.Value(loose[0]))
+
+# The comparison 0 < cr1 <= cr2
+bounded_reg = expr.And(
+    expr.Less(expr.Value(0), expr.Value(cr1),
+    expr.LessEqual(expr.Value(cr1), expr.Value(cr2),
+)
+```
+
+These expressions will then be given directly to `IfElseOp` or `WhileLoopOp` in their `condition` fields.
```
kdk, line 113: For sake of clarity, it might help to show these examples in the context of building an `IfElseOp` or `WhileLoopOp` since it seems a more common usage than building expressions separately.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+These expressions will not be valid in the general `Instruction.condition` field; we do not want to expand any support for something we are already moving to remove.
+
+
+#### Stage 2
+
+Looking beyond the initial phase, it will be convenient to define a less verbose way for users to create these expressions.
+Since this document is intended to define a representation that can be removed and deprecated from Terra easily if it needs to evolve, we will _not_ consider overloading the Python magic-methods `__and__` etc on `Clbit` and `ClassicalRegister` in general Terra usage.
+However, we can do so in a limited scope, which will temporarily change the semantics.
+This will look like (continued from above code block, and the same examples in the same order):
```
kdk, line 122: I think it's worth evaluating how much this makes the deprecation and eventual removal path smoother for users vs. the near time usability cost. It seems like most usage of `IfElseOp` and `WhileLoopOp` will want a `with expr.build():` block, and (with the exception of `__eq__`) these magic methods aren't currently being used, and so could be deprecated as needed when used.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+
+```python
+with expr.build():
+    equal_reg = cr1 == 2
+    equal_bit = loose[0] == False
+    not_bit = ~loose[0]
+    bounded_reg = (0 < cr1) & (cr1 <= cr2)
+```
+
+It is not possible to overload the behaviour of the Python Boolean operators (`not`, `and` and `or`) for classes, which is why the bitwise operations `~` and `&` are used instead (with the corresponding precedence frustrations).
```
kdk, line 132: I could see this being a source of confusion. e.g. "Why can I `cr[0] == cr[1]` but not `cr[0] and cr[1]`?". Are there any cases where this could lead to errors like "I wanted a bitwise AND but got a logical AND"?

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+Within the `expr.build()` context, the magic methods of `ClassicalRegister` and `Clbit` will be temporarily overridden to be an eager builder interface for the syntax tree given above.
+This means that the expressions will not need to be bound to variables in order to use them, as the resulting objects will be well-defined and the correct tree form.
```
kdk, line 134: I agree with the statement here, but I'm not sure how it follows from the sentence above.

kdk (commented):
Looks good so far. Thanks for writing this up.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+The class `expr.Expression` will be an abstract base class with two read-only fields:
+```python
+# file: qiskit/circuit/classical/expr.py
+
+class Expression(ABC):
+    resources: typing.Set[Clbit]
+    type: qiskit.circuit.classical.Type
+```
+
+The `resources` field is effectively a `use-def` tracker, which will more easily enable the conversion to data-flow (`DAGCircuit`) format to evaluate the necessary wires of an instruction.
```
kdk, line 164: Not a compiler expert by any means, but I think this is not exactly the same as `use-def` tracking, which makes some assurances about the value of a variable not being re-defined between definition and use. The `Clbit`s in `QuantumCircuit` now follow more like memory semantics, so there are no guarantees of their value or when they were last modified.

jakelishman: Yeah, you're right.  In retrospect, I'm not even going to have the `resources` field anyway - I'll just make an `ExpressionVisitor` subclass that calculates them.  There's no reason that these "resources" should be any more special than any other shared property.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+- This support is not limited to IBM so may extend further, but we are not aware of other dynamic-circuits hardware accessible through Qiskit
+- For Terra development: this is a "build one to throw away"-type sytem, so we can see in actual usage what works and doesn't for us before we commit to further classical processing.
+
+
+## Design Proposal
+
+The intent of this RFC is to define an exploratory representation for a simple subset of classical bit-level operations, which expand the current state of Terra's `IfElseOp.condition`.
+This representation is not required to be the base for future expansion; it will be permissible to introduce a completely different classical-value type system if that is more appropriate, and deprecate this one.
+To that end, this proposal will not include adding any additional public methods to any existing Terra classes that might conflict with an alternative implementation.
+In particular, this proposal will not consider adding ease-of-use arithmetic methods to `Clbit` or `ClassicalRegister`, nor will it add expressions that can be used as the parameters to gates.
```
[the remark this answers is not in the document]
jakelishman: I can add a couple of lines about that if you like, though there's equally no hard guarantee that it _will_ be removed; I was just designing such that it could be our standard methods of deprecation.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+  - `bool -> bool`: `!` (boolean negation).
+- Binary:
+  - `uint[n] * uint[n] -> uint[n]`: `&`, `|`, `^`;
+  - `bool * bool -> bool`: `&&`, `||`;
+  - `uint[n] * uint[n] -> bool`: `==`, `!=`, `<`, `<=`, `>`, `>=`.
+
+For typing in this initial draft, we will make an exact bijection:
+
+- `ClassicalRegister(size=n)` <=> `uint[n]`
+- `Clbit` <=> `bool`
```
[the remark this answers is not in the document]
jakelishman: `Clbit` won't be a `uint[1]` in this type system, nor I think should it be eligible for any implicit cast to `uint[1]`, but I will overload `&`, `|` and `^` to also act on `bool` too (I actually thought that's what I'd written, but apparently not).

When we add explicit casting, I'm fine having an explicit cast `bool -> uint[_]` defined by `true -> 1` and `false -> 0`, but having a special implicit cast from `bool -> uint[1]` I think is liable to cause more problems than it solves, and I think all the real benefit can be derived by just allowing the bitwise operators to act on them.

0030-simple-classical-representations.md
jakelishman: When I wrote this, I was intending `Equal(Value(2), Value(2))` to be a type error on construction, because that expression would involve evaluating a binary op with both inputs having type `uint[UNKNOWN]`, and that's something I wanted to avoid having to define semantics for.

For the inference, this is what I meant:
```python
creg = Value(ClassicalRegister(3))  # type: uint[3]
two = Value(2)  # type: uint[UNKNOWN]
three = Value(3)  # type: uint[UNKNOWN]
five = Value(5)  # type: uint[UNKNOWN]

# This binary op causes the type of `two` to be inferred as `uint[3]` to match.
# This would recurse through any partially type objects.
add_two = Add(creg, two)

# This expression is not valid, because it's a binop on two partial types.
eight = Add(three, five)
```

My intent was just to forbid this kind of thing for now in the actual expression tree, rather than needing to define semantics for how we handle intermediate numeric values of unknown width.  (For example, at what precision would we choose to handle an expression `312890318904839189082032 < 43819432893121004381904`?)

0030-simple-classical-representations.md
jakelishman: If we choose to keep these objects, we can also just move the magic methods onto the core objects, but that doesn't really feel great to me; I think magic methods should always feel _obvious_ about what they're going to do, and detached from a builder context, I don't think "`Clbit == Clbit` makes a runtime expression" is super obvious.

0030-simple-classical-representations.md
jakelishman: The way I wrote this, you'll always have the option to use the regular constructors, or potentially Kevin's suggested chained methods, even in the middle of the builder context.  But yeah, it's not ideal, but imo magic methods give the best expression legibility in the majority of cases, so until we decide to commit to a _complete_ function-based builder interface where we can do AST inspection, this is my preferred choice.

Also: my intent was for `&` to still be bitwise AND, it's just also defined on bools.  You'd have to be explicit to get logical operations (with the corresponding short-circuiting)

0030-simple-classical-representations.md
jakelishman: I was trying to say that you don't need to do
```python
with expr.build():
   my_cond = creg1 == creg2
with qc.if_test(my_cond):
   pass
```
(i.e. you don't need the `build` context to end), you'd be able to directly build the circuit as
```python
with expr.build():
    with qc.if_test(creg1 == creg2):
        pass
```
and the `if_test` method would see the correct objects at the point of call.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,260 @@
...
+The allowed expressions will be the following:
+
+- Object:
+  - literal `bool` (`True` or `False`);
+  - literal `uint[_]` (non-negative Python `int`);
+  - runtime `bool` (`Clbit`);
+  - runtime `uint[n]` (`ClassicalRegister(size=n)`).
+- Unary:
+  - `uint[n] -> uint[n]`: `~` (bitwise negation);
+  - `bool -> bool`: `!` (boolean negation).
```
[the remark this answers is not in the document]
jakelishman: Oh, that was a dumb C/Rust/Python mixup.  I guess we'd have to rely on `~` for logical negation too.

0030-simple-classical-representations.md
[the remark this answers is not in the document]
jakelishman: Added a note about not requiring this to be removed and a short note on deprecations in e6dd506.

0030-simple-classical-representations.md
jakelishman: Done in 7f5424e.

0030-simple-classical-representations.md
jakelishman: Fixed in 2f80715.

0030-simple-classical-representations.md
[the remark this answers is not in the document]
jakelishman: Fixed in b8b7a19.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,319 @@
...
+- Object:
+  - literal `bool` (`True` or `False`);
+  - literal `uint[_]` (non-negative Python `int`);
+  - runtime `bool` (`Clbit`);
+  - runtime `uint[n]` (`ClassicalRegister(size=n)`).
+- Unary:
+  - `uint[n] -> uint[n]`: bitwise negation
+  - `bool -> bool`: Boolean negation
+- Binary:
+  - `uint[n] * uint[n] -> uint[n]`: `&`, `|`, `^`;
```
ewinston, line 60: Should bit shift `<<` and `>>` be allowed as well?

jakelishman: For sure, in the future - this is just meant to be an MVP and I didn't want the available expression tree to be huge for a first attempt.

kdk: I think given that these are already defined in OpenQASM3 (and available on IBM hardware), and open the door to some applications, it would be good to have these in the first MVP (unless there's a subtlety that would make them more complex to implement than those already listed).

0030-simple-classical-representations.md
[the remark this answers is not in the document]
kdk: This makes sense to me, can you add a note on this to future extensions (and update to include the overloading for `Clbit`s)?

0030-simple-classical-representations.md
jakelishman: Yeah, that would raise or warn becuase `0b1010` can't fit in a `uint[3]`.

0030-simple-classical-representations.md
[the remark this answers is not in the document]
jakelishman: Done in 041bf5d.

0030-simple-classical-representations.md
kdk: That's a good point, and agree that's a good approach for now.

kdk (approved):
Thanks for the updates, this LGTM.

0030-simple-classical-representations.md
```diff
@@ -0,0 +1,319 @@
+# Preliminary representation of rvalue classical expression in Qiskit
+
+| **Status**        | **Proposed/Accepted/Deprecated** |
+|:------------------|:---------------------------------------------|
+| **RFC #**         | ####                                         |
```
1ucian0, line 5:
```suggestion
| **Status**        | **Accepted** |
|:------------------|:---------------------------------------------|
| **RFC #**         |  30                                        |
```

PR #34, opened 2023-04-26
Qiskit release cycle and versioning

1ucian0:
A new versioning schema for a more clear stability cycle. SemVer with CalVer elements.


garrison:
My first impression is that I think that bumping the major version once a year sounds about right.   If semantic versioning is followed, this means only introducing breaking changes once per year.  I could see two workflows that could support this.

Option 1:

- Both major versions and minor versions are branched from `main` shortly before release.
- New features can be merged anytime throughout the year, which then appear in the next minor release.  However, breaking changes can only be merged in a 2-3 month window that occurs once per year.  In the worst case, breaking changes must wait up to ~10 months before they can merge, ~12 months before they can be released.

Option 2:

- Major versions are branched from `main` shortly before release.  Minor releases are then branched from the most recent _major_ release branch shortly before that minor release.
- Breaking changes can be merged to `main` at any time throughout the year.
- All non-breaking, new features need to be backported to the most recent major release branch right after the merge to `main`.

I prefer Option 1, as it means fewer branches and directs human effort and eyeballs toward the next minor release.  Breaking changes can be difficult and annoying to backport, too.  But I fear that there will be pressure to hold up each major release for "just one more breaking change" with a release cycle so long.  Especially in that case, the deadline for having a PR ready must be well before the major version's release date.  Just as Linus Torvalds hates when merge requests come in at the end of the `rc1` window, I could see this leading to some significant pressure on the Qiskit maintainers.

Having thought through this more thoroughly, **would we be better off to start with a 6-month cycle for a bit?**  One major release, then one minor release 3 months later, then back to major again 3 months after that, etc.  Many, many open source projects -- from desktop environments to entire distribtutions -- have found 6 months to be the correct cadence.

A few remaining thoughts/questions:

- `DeprecationWarning`s should not be introduced in a minor release if semantic versioning is being followed.  I _think_ it would be fine for a minor version to introduce a `PendingDeprecationWarning`, though, which would then become a `DeprecationWarning` in the following _major_ release.
- I don't fully understand the bars in the mermaid plots.  Do they represent the fraction of time that a branch should expect to receive stable backports?

jakelishman:
I'm with Jim above.  I think his option 1 is probably the right way for us to try, at first at least - I don't think we have sufficient manpower or dev-team experience to attempt to manage development of two versions simultaneously (which is kind of what option 2 devolves to without a lot of effort).  I'd also be happy to try a six-month cadence initially (also content to stick with yearly), though in this scenario we'd only have one minor release between majors, so I'm not certain if it'd actually buy us _too_ much.

Just to speak to the penultimate point: the formal semantic versioning spec allows deprecations to be added in minor releases ([link to rule](https://semver.org/#spec-item-7), [link to faq entry](https://semver.org/#how-should-i-handle-deprecating-functionality)).  To keep the spirit of our current deprecation policy, I think we should still maintain the rule "deprecation warnings can only occur 1+ minor releases after the replacement functionality is added", and potentially strengthen the removal rule to "deprecation warnings must be emitted for a complete major version" (i.e. 1+ year).

####-release_cycle.md
```diff
@@ -0,0 +1,138 @@
...
+>   - If spans multiple projects cover these parts individually
+> - Interaction with other features
+> - Dissecting corner cases
+> - Reference definition, eg., formal definitions.
+
+
+### Deprecations
+
+* deprecation warnings can only occur 1+ minor releases after the replacement functionality is added
+* deprecation warnings must be emitted for a complete major version" (i.e. 1+ year)
```
1ucian0, line 123: Added from https://github.com/Qiskit/RFCs/pull/34#issuecomment-1549428294 @jakelishman 

I would like to offer an alternative as this sounds like a lot of time to me. Introducing a deprecation warning in X.Y and not being able to remove it until X+2.0 is a huge time period of dangling code. If we keep X.3 support for extra 6 month, that should be enough to remove it in X+1.0.

1ucian0:
> I don't fully understand the bars in the mermaid plots. Do they represent the fraction of time that a branch should expect to receive stable backports?

The plot represent how long version has support. X.1 is supported until X.2 is released, for example.

####-release_cycle.md
jakelishman: I see the trouble with this approach is that it assumes that every user will upgrade through every minor version, so will see deprecations introduced in minor releases.  To me, introducing major versions means that we expect that the required upgrade path is only at the granularity of "through each major version", and so my form is what guarantees that a user will see every deprecation warning.

jakelishman: For example, currently our upgrade strategy is that you can't expect to go from Terra 0.22 to 0.24 without something breaking without warning - you must go via 0.23 to ensure you've seen all warnings and had the opportunity to smoothly upgrade.  I would foresee that if we add major versioning, the story ought to be "you must upgrade through each major, but you needn't install each minor".

garrison: > it assumes that every user will upgrade through every minor version, so will see deprecations introduced in minor releases.

Actually, the necessary assumption is weaker: before upgrading to a new major version, the user would only need to use the _final_ minor version of the previous release, because there, all deprecation warnings necessary for upgrading will be visible and can be fixed.

Julia is an example that illustrates both the benefits and drawbacks of this approach.  Julia 0.7.0 and 1.0.0 were released within ~48 hours of each other.  The only difference between them is that 0.7.0 contained compatibility (and deprecation warnings) on code that worked under 0.6, while 1.0 removed all such support.  If there is ever a Julia 2.0, I expect the same thing to happen: users will be expected to use the "final" 1.X release and fix all deprecation warnings before trying the code under Julia 2.0.

The downside to this approach was that, in practice, people not intimately familiar with the release plan upgraded directly from 0.6.x to 1.0, and were met with a poor user experience: tons of code that worked on the previous release simply failed.  If the goal is to get users to use a certain release before upgrading, it really does need to be emphasized in _all_ the obvious locations, and then some.

1ucian0: > people not intimately familiar with the release plan upgraded directly from 0.6.x to 1.0, and were met with a poor user experience: tons of code that worked on the previous release simply failed.

I see this as expected, as a major upgrade.

> I see the trouble with this approach is that it assumes that every user will upgrade through every minor version

Indeed. This is an assumption that needs to be explicit. However, it is not more obscure than the current one.
Downstream library should depend on the fairly standard `~=X.Y` or `==X.*`. With the extended period for `X.3.*` receiving critical bug fix support and not introducing new deprecations, pinning the previous major is always an option and gives a good period of time to migrate.

####-release_cycle.md
```diff
@@ -0,0 +1,132 @@
...
+    X+2.0    :rX2Y0, after rX1Y3, 91d
+    X+2.1    :rX2Y1, after rX2Y0, 91d
+    X+2.2    :rX2Y2, after rX2Y1, 91d
+    X+2.3    :rX2Y3, after rX2Y2, 91d
+    X+2.3.*  :done, rX2Y3s, after rX2Y3, 183d
+```
+
+## Detailed Design
+
+> Technical reference level design. Elaborate on details such as:
```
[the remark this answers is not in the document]
1ucian0: > Removals/breaking changes can only happen during one 3 month period of the year (the cycle leading up to the release of a new major version).

Not sure  what "can only happen" means in this context. If Option 1 from https://github.com/Qiskit/RFCs/pull/34#issuecomment-1548520331 (which i agree with), then can merge in main at any point. 

>Following https://github.com/Qiskit/RFCs/pull/34#issuecomment-1549428294, possibly an extended time deprecations would be shown before being candidates for removal.

In the currently proposal, that number is a minimum of 6 month. I think this is an extension on the current one (3 months - one full minor). Added here: 5e56c5136f1c22e78fc22f515fc40546a04cac42

> Extended timeline for back porting bug fixes to prior releases. (Currently, this is only the most recent release, but do we have an idea how far this would go under the proposal?)

The current suggestion is to the current stable with support, which is 1.5 years.

####-release_cycle.md
jakelishman: My issue is that adding major versions into the mix does change the "obscurity" of the policy.  I expect that code I write for one version of a library with no warnings will work in the next version of that library (now maybe with warnings), at whatever granularity that library does it major releases.  Once we have majors, I see them as defining the required-upgrade granularity, as opposed to our current system where minors are that top level.  That's not a requirement of semantic versioning as a spec, but that's because that spec is only for associating the scope of changes with the version number, not for any of the development/upgrade strategies that go with it.

Requiring that the 2->3 upgrade goes via 2.last and heavily publicising that is one way of dealing with my concern, but I think we've seen that we won't ever reach all our users, and I'd argue that this is a much more "obscure" requirement.  People expect instability when something's versioned 0.x, but there's more expectation of stability when it's 1.x+.

blakejohnson: @garrison I actually really liked the Julia approach of having a final minor version to introduce all the deprecations prior to the major version bump. It provided a concrete path to follow as a developer to get to working code despite introducing a lot of change in 1.0.

In general, I think relying upon the _final_ minor version as a necessary step to expose all the deprecations is a very teachable system.

blakejohnson:
What if instead of committing to releasing a new major version each year, we simply commit to updating which version is considered "stable"? That decouples semvar signaling of version compatibility from a statement about stability / support.

wshanks:
A couple things I think it would be good to include (though it seems like the draft might change a lot based on the comments above):

1. The git graph for the proposed new branching model on shows features and bugfixes. Where do the breaking changes happen? Is it that all breaking changes are removals guarded by a deprecation warning and then removed the `ready_for_plus_one` commit? There could be behavioral changes but maybe those would be forced to be an addition of a new name plus a removal of an old one, rather than a change in behavior of one function or method (not sure if that is always possible). Partly I am wondering how one tests the next version of qiskit without an extra branch with the breaking changes. Maybe they get applied after X.3 and then can be tested between then and X+1.0.
2. There is open discussion about the support cycle above, so the current proposal may not be followed any way, but I think it might be worth including some guidance for downstream projects. For example, currently, the IBM client packages `qiskit-ibm-provider` and `qiskit-ibm-runtime` tend to bump the minimum version of `qiskit`(-terra) they depend on soon after there is a new `qiskit` release. Since these clients also include changes necessary to keep working with the IBM Quantum API, it is not viable to hold back on an old version of the client and so anyone working with IBM systems has to upgrade `qiskit` promptly as part of the client upgrade. So at least for those working with IBM systems, there would not be much value to providing a long support cycle for X.3 after X+1.0 is released if things keep working as they have been.

1ucian0:
I think all your comments are addressed. Have a look!

####-release_cycle.md
```diff
@@ -0,0 +1,325 @@
...
+| **Updated**       | YYYY-MM-DD                                   |
+
+## Summary
+
+Balancing stable backward compatibility with the rapid pace of technology is a crucial consideration for Qiskit. This RFC is a public discussion that aims to find a harmonious solution that offers users a clear and transparent support contract while minimizing the burden on developers and users. Ideally, we are in  the search for an approach that also maximizes flexibility to accommodate new hardware and services, which often necessitate the development of new features.
+
+
+## Motivation
+
+Qiskit is known for its dynamic nature and frequent deprecation of features. However, it [deprecation policy](https://qiskit.org/documentation/deprecation_policy.html) places a strong emphasis on stability and guarantees a relatively lengthy transition period (usually longer than six months from the first notification point on) when removing a feature. For the last years, Qiskit also introduced planned releases for mitigating the impact of coming changes, as well as pre-releases.
```
mtreinish, line 17:
```suggestion
Qiskit is known for its dynamic nature and frequent deprecation of features. However, the [deprecation policy](https://qiskit.org/documentation/deprecation_policy.html) places a strong emphasis on stability and guarantees a relatively lengthy transition period (usually longer than six months from the first notification point on) when removing a feature. For the past several years Qiskit has also had planned releases on a regular schedule for mitigating the impact of coming changes, as well as pre-releases to enable testing prior to a final release.
```

####-release_cycle.md
```diff
@@ -0,0 +1,325 @@
...
+
+However, there is a demand for a more transparent release cycle with longer periods of backwards compatibility support. In order to get ready for a "beyond Qiskit 0.x", a discussion on how that cycle would like is key to understand the trade-offs among:
+
+ * developer effort: Supporting several stable versions requires significant development resources. At the same time, rolling releases tend to create technical debt and few chance of rebuild from scratch particular modules.
```
mtreinish, line 21:
```suggestion
 * developer effort: Supporting several stable versions requires significant development resources. At the same time, rolling releases tend to create technical debt and fewer chances to enable re-architecting features or interfaces for better maintainability or changing needs.
```

####-release_cycle.md
```diff
@@ -0,0 +1,325 @@
...
+
+
+## Motivation
+
+Qiskit is known for its dynamic nature and frequent deprecation of features. However, it [deprecation policy](https://qiskit.org/documentation/deprecation_policy.html) places a strong emphasis on stability and guarantees a relatively lengthy transition period (usually longer than six months from the first notification point on) when removing a feature. For the last years, Qiskit also introduced planned releases for mitigating the impact of coming changes, as well as pre-releases.
+
+However, there is a demand for a more transparent release cycle with longer periods of backwards compatibility support. In order to get ready for a "beyond Qiskit 0.x", a discussion on how that cycle would like is key to understand the trade-offs among:
+
+ * developer effort: Supporting several stable versions requires significant development resources. At the same time, rolling releases tend to create technical debt and few chance of rebuild from scratch particular modules. 
+ * user support: users tend to demand longer support periods to avoid regular updates on their code and software.
```
mtreinish, line 22:
```suggestion
 * user support: users tend to demand longer support periods to avoid regular updates in their code and software.
```

####-release_cycle.md
```diff
@@ -0,0 +1,325 @@
...
+ * new feature support for coming technology: the quantum computing field and hardware is in constant change and scaling to it many time requires big changes in Qiskit, not always compatible with previous approaches.
```
mtreinish, line 23:
```suggestion
 * new feature support for coming technology: the quantum computing field and hardware is constantly changing and scaling up. This can require changes in Qiskit,which are not always compatible with previous approaches.
```

####-release_cycle.md
```diff
@@ -0,0 +1,325 @@
...
+    commit id:"bugfix/3"
+    checkout stable/0.X
+    cherry-pick id:"bugfix/3"
+    checkout main
+    commit id:"feature/4"
+    checkout stable/0.X
+    commit id: "tag_0.X.1" tag: "v0.X.1"
+```
+
+The main branch is a single development branch from which some bugfixes are ported to the stable branch, from which releases are done.
```
mtreinish, line 106:
```suggestion
The main branch is a single development branch from which some bugfixes are backported to the stable branch, from which releases are done.
```

####-release_cycle.md
```diff
@@ -0,0 +1,325 @@
...
+
+
+The new release cycle could include:
```
mtreinish, line 109:
```suggestion
The new release cycle will include:
```

####-release_cycle.md
```diff
@@ -0,0 +1,325 @@
...
+
+ * [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html)
+ * Critical bugfix support for a year or longer
```
mtreinish, line 112:
```suggestion
 * Critical bugfix support for at least one year

File: ####-release_cycle.md [hunk 10]
```diff
@@ -0,0 +1,325 @@
...
+ * 6-month transition period for updating to a new major while having support.
+ * Non-breaking new features in a 3-month cycle (as currently)
```
mtreinish, line 114:
```suggestion
 * Non-breaking new features releases in a 3-month cycle (as currently)
```

####-release_cycle.md
```diff
@@ -0,0 +1,325 @@
...
+    X+1.2    :rX1Y2, after rX1Y1, 91d
+    X+1.3    :done, rX1Y3, after rX1Y2, 274d
+    section X+2.*
+    X+2.0    :rX2Y0, 2026-01-31, 91d
+    X+2.1    :rX2Y1, after rX2Y0, 91d
+    X+2.2    :rX2Y2, after rX2Y1, 91d
+    X+2.3    :done, rX2Y3s, after rX2Y2, 274d
+```
+
+Notice that i.3 has an extra 6 month bug fixing support. There is no i.4, since it coincides with i+1.0, where new features (and possibly breaking changes) are released.
```
mtreinish, line 140:
```suggestion
```mermaid
gantt
    dateFormat  YYYY-MM-DD
    tickInterval 3month
    todayMarker off
    axisFormat %b
    section X.*
    X.0   :rX0Y0, 2024-01-31, 91d
    X.1   :rX0Y1, after rX0Y0, 91d
    X.2   :rX0Y2, after rX0Y1, 91d
    X.3   :rX0Y3, after rX0Y2, 91d
    X.4,  :done, rX0Y4, after rX0Y3, 183d
    section X+1.*
    X+1.0    :rX1Y0, 2025-01-31, 91d
    X+1.1    :rX1Y1, after rX1Y0, 91d
    X+1.2    :rX1Y2, after rX1Y1, 91d
    X+1.3    : rX1Y3 after X1Y2, 91d
    X+1.4,   :done, rX1Y4, after rX1Y3, 183d
    section X+2.*
    X+2.0    :rX2Y0, 2026-01-31, 91d
    X+2.1    :rX2Y1, after rX2Y0, 91d
    X+2.2    :rX2Y2, after rX2Y1, 91d
    X+2.3    :rX2Y3s, after rX2Y2, 91d
    X+1.4,   :done, rX2Y4, after rX2Y3, 183d
```
Notice that i.4 has an extra 3 months of bug fixing support. The i.4 releases have no API changes and are identical to the i.3 releases and just includes additional deprecation warnings to prepare for the potentially breaking API changes in i+1.0.
```

**Review by mtreinish — APPROVED** [review]
LGTM, just some small things inline. I think we can merge after those corrections.

File: ####-release_cycle.md (hunk 1 above)
Comment (1ucian0) [reply ####-release_cycle.md]: The notion of "final minor release" was introduced in 3ec912ba0ae7e5664cdc03692f45ac9506caefea

File: ####-release_cycle.md [hunk 12]
```diff
@@ -0,0 +1,343 @@
...
+All changes, including bug fixes, with backward compatibility are PRed with `main` as base. Bugfix PRs that are possible to port to stable should be tagged as `stable-backport-potential` and a bot ports them to `stable/X.Y` (where `Y` is the last minor).
+At undefined time, a new patch release `X.Y.Z` is tagged on `stable/X.Y` and released.
+
+Non-backward compatible changes, including removal of deprecated code, can be PRed at any point against `next/X+1` branch.
+
+Regularly, `main` should be merge into `next/X+1` to keep it up-to-date. The suggestion is for every minor release, at least.
+
+### QPY support
+
+A new major release `X` has support to dump on QPY in the same version than the final minor version `X-1`. This allows `X` to dump QPY that can be loaded with `X-1`. Minor releases can increase the version of QPY but should always support the QPY version in the final minor release of `X-1` until the end-of-life of `X-1`.
```
mtreinish, line 308: Maybe something like this?: 

```suggestion
The QPY serialization format is backwards compatible so that a new Qiskit can release can always load a QPY
file generated with an earlier release of Qiskit. However, the format isn't forward compatible, so it's not possible
to load QPY files generated with newer version of Qiskit using an older release. To support compatibility across major version releases, the `qiskit.qpy.dump()` function has an optional argument `version` that lets a user specify a back version of the QPY format emitted by the `dump()` function. There are two constants exposed by
the `qpy` module, `qiskit.qpy.QPY_VERSION` which documents the default format version emitted by `dump()` and
also the maximum version that `load()` supports, and `qiskit.qpy.QPY_COMPATIBILITY_VERSION` which documents the minimum version that `dump()` can emit (while `load()`  can load any version <= `QPY_VERSION`).

To facilitate user migration across major version releases the `qiskit.qpy.dump()` function will always support at least one overlapping version between the `X.0.0` and the `X-1.Y.0` release (where Y is the last minor version of 
that series). This will enable saving QPY format files that can be loaded by both major versions from the newer
release. For example, in Qiskit 1.0.0 the `QPY_VERSION` was `11`, and in Qiskit 0.46.0 the `QPY_VERSION` constant was `10`. For qiskit 1.0.0 the minimum version supported by `dump()`'s `version` argument, and exposed by 
`qiskit.qpy.QPY_COMPATIBILITY_VERSION` was 10. This enabled Qiskit 1.0.0 to generate qpy payloads that could
be loaded by Qiskit 0.46.0.
```

PR #38, opened 2023-04-27
Sequential numbering for 0012, 0025, and 0030

1ucian0:
The RFC process defines a sequential numbering. So:
- renamed:    0012-unify-pulse-commands-and-instructions.md -> 0008-unify-pulse-commands-and-instructions.md
- renamed:    0025-interface-for-circuit-operations.md -> 0009-interface-for-circuit-operations.md
- renamed:    0030-simple-classical-representations.md -> 0010-simple-classical-representations.md

### DIFF
--- a/0008-unify-pulse-commands-and-instructions.md
+++ b/0008-unify-pulse-commands-and-instructions.md
@@ -2,7 +2,7 @@
 
 | **Status**        | **Proposed**             |
 |:------------------|:---------------------------------------------|
-| **RFC #**         | 0012                                         |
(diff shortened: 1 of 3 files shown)


jakelishman:
"Sequential" doesn't necessarily require "consecutive" - I think it's better to keep the relation of an RFC to its historical context via the PR review.

1ucian0:
Yeap. I should had said _consecutive_. Alternatively, we could link to the PR in the README table?

PR #31, opened 2023-03-31
Plan to rename `Qiskit/qiskit-terra` repo to `Qiskit/qiskit`

1ucian0:
This RFC is about planning the `Qiskit/qiskit-terra` -> `Qiskit/qiskit` rename. It will be open for comments and feedback for 10 days (until April 10th).

PR #39, opened 2023-04-27
Numbering RFC `0011`

1ucian0:
The RFC titled _Plan to rename `Qiskit/qiskit-terra` repo to `Qiskit/qiskit`_ is assigned number 0011.

### DIFF
--- a/0011-repo-rename.md
+++ b/0011-repo-rename.md
@@ -2,10 +2,10 @@
 
 | **Status**        | **Proposed** |
 |:------------------|:---------------------------------------------|
-| **RFC #**         | ####                                         |
+| **RFC #**         | 0011                                         |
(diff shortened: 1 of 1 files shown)


jakelishman (commented):
I think it would be useful to associate the RFC number with the PR number that added it, so the discussion surrounding the PR and any historical changes to it are stored.  In that case, we might want to assign this RFC number #31.

1ucian0:
If I remember @mtreinish 's Readme interpretation, the idea is to number the rfcs after merge, in sequential order. Maybe something to discuss?

1ucian0:
> I think maintaining a link to the PR history is probably a good idea, in the spirit of the original intent.

done in 8411ffb83118fe1b3713a6e063cb3177c10e39df Thanks!

mtreinish (approved):

PR #44, opened 2023-06-27
qiskit.algorithms move as an independent package

ElePT:
In an attempt to inform and coordinate the move of the algorithms repo, I have quickly pulled together an RFC.

### DIFF
--- a/0012-algorithms-migration.md
+++ b/0012-algorithms-migration.md
@@ -14,7 +14,7 @@ This RFC was created to document discussions concerning the future of the `qiski
 This module provides a series of interfaces and implementations of commonly used quantum algorithms based on 
 Qiskit's core components. Historically, the module was created as part of Qiskit Aqua, together with the now
 community-based application modules (Nature, Machine Learning, Optimization, Finance). 
-It [currently "lives" in `qiskit-terra`](https://github.ibm.com/IBM-Q-General/user-feedback/issues),  
+It [currently "lives" in `qiskit-terra`](https://github.com/Qiskit/qiskit),  
 and over the past year has been refactored to use the Qiskit Primitives and sustain the 
 Opflow and Quantum Instance deprecations.
 
--- a/0012-algorithms-migration.md
+++ b/0012-algorithms-migration.md
@@ -13,8 +13,7 @@
 This RFC was created to document the discussions around the future of the `qiskit.algorithms` module.
 This module provides a series of interfaces and implementations of commonly used quantum algorithms based on 
 Qiskit's core components. Historically, the module was created as part of Qiskit Aqua, together with the now
-community-based application modules (Nature, Machine Learning, Optimization, Finance). It currently "lives"
-in `qiskit-terra`, and over the past year has been refactored to use the Qiskit Primitives and sustain the 
+community-based application modules (Nature, Machine Learning, Optimization, Finance). It [currently "lives" in `qiskit-terra`](https://github.ibm.com/IBM-Q-General/user-feedback/issues), and over the past year has been refactored to use the Qiskit Primitives and sustain the 
 Opflow and Quantum Instance deprecations.
 
 The content of `qiskit.algorithms` is heterogeneous. Some components are algorithm implementations, while others
--- a/0012-algorithms-migration.md
+++ b/0012-algorithms-migration.md
@@ -4,34 +4,83 @@
 |:------------------|:------------------------------------------------------------------------------------------|
 | **RFC #**         | ####                                                                                      |
 | **Authors**       | [Elena Peña Tapia](https://github.com/ElePT), [Steve Wood](https://github.com/woodsp-ibm) |
-| **Submitted**     | 2023-06-27                                                                                |
+| **Submitted**     | 2023-07-05                                                                                |
 | **Updated**       | YYYY-MM-DD                                                                                |
 
 
 ## Summary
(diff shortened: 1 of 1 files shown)


ElePT:
> Thanks! Where will the docs move to? qiskit.org/ecosystem/algorithms? That name would match how we have e.g. `qiskit-nature` at `ecosystem/nature`.
> 
> FYI we can set up redirects in cloudflare from the qiskit.org/documentation/ subpages to ecosystem/algorithms if we want.

Exactly, I was thinking of `qiskit.org/ecosystem/algorithms`. And I was just about to add the docs redirect idea to the RFC when I read your comment, I think it makes a lot of sense!

kdk (approved):
Following discussion in the team meeting today, I'm good with path 1.

0012-algorithms-migration.md
```diff
@@ -0,0 +1,178 @@
...
+| **Updated**       | YYYY-MM-DD                                                                                |
+
+
+## Summary
+
+This RFC was created to document the discussions around the future of the `qiskit.algorithms` module.
+This module provides a series of interfaces and implementations of commonly used quantum algorithms based on 
+Qiskit's core components. Historically, the module was created as part of Qiskit Aqua, together with the now
+community-based application modules (Nature, Machine Learning, Optimization, Finance). It currently "lives"
+in `qiskit-terra`, and over the past year has been refactored to use the Qiskit Primitives and sustain the
```
1ucian0, line 17:
```suggestion
Qiskit's core components. Historically, the module was created as part of Qiskit Aqua, together with the now
community-based application modules (Nature, Machine Learning, Optimization, Finance). It [currently "lives" in `qiskit-terra`](https://github.ibm.com/IBM-Q-General/user-feedback/issues), and over the past year has been refactored to use the Qiskit Primitives and sustain the 
```

0012-algorithms-migration.md
```diff
@@ -0,0 +1,178 @@
...
+Opflow and Quantum Instance deprecations.
+
+The content of `qiskit.algorithms` is heterogeneous. Some components are algorithm implementations, while others
+are "sub-routines" or algorithm utilities, such as gradients, optimizers or state fidelities. Some of these
+utilities are implementatons from scratch, others are wrappers over commonly used libraries (such as the 
+SciPy optimizers).
+
+With Qiskit moving towards a leaner definition, we are now revisiting the future of this module. 
+We would like to tell apart the fundamental components for algorithm design from algorithm implementations, and 
+find a better suiting location for both of them while minimizing the disruption caused to users.
```
1ucian0, line 27:
```suggestion
With Qiskit moving towards a leaner definition and focusing on primitives as quantum hardware access model, the future of `qiskit.algorithms` is being revisited. The classes in the module tend to hide the primitives interfaces and serve as convenience wrappers for abstractions, beyond the current definition of Qiskit.
As such, they do provide value to users to it would be important to 
find a better suiting location for them while minimizing the disruption caused to users.
```

0012-algorithms-migration.md
```diff
@@ -0,0 +1,182 @@
...
+
+**Repos immediately affected by the plan**: `Qiskit/qiskit-terra` and `Qiskit/qiskit-tutorials`
+
+### qiskit-terra
+
+1. Open a PR to add a single deprecation message telling users to change their imports 
+from `qiskit.algorithms` to `qiskit_algorithms`. The code will continue existing 
+in `qiskit-terra` until the deprecation period is over [Blocker: release of `qiskit_algorithms-0.1.0`,
+see below].
+2. What should we do with the algorithms migration guide? we could leave it in terra during the deprecation period,
```
1ucian0, line 105: Migration guides usually stay online after the deprecation period.

ElePT: This particular guide is for migrating from QuantumInstance-based `qiskit.algorithms` to primitive-based `qiskit.algorithms`. I am all for keeping guides for as long as possible, but I am not sure how to balance them with "completely decoupling qiskit.algorithms from qiskit" efforts. That is why we suggested to move this guide to the `qiskit-algorithms` repository (but keep the opflow/quantum instance ones in qiskit). It would stay alive but decoupled.

mtreinish: At least when I remember discussing this a while ago when @woodsp-ibm brought this up the thinking was that we'll have the stable docs builds for historical releases so we can just point people to: https://qiskit.org/documentation/stable/0.44/migration_guides/algorithms_migration.html and not worry about having to keep the guides in source form forever.

woodsp-ibm: As we do not have older versions of docs published in the apps what we have been doing there, once a migration guide no longer can be run/built from main is to convert it as a rst file so it gets included but its content is no longer executed. I had been thinking that if the guide was wanted to be kept/publihsed longer in Qiskit that potentially in main its contents just got replaced by a link to the stable published version where it existed since the older versions of the docs were kept. That seemed simpler to me and also being a link one could state that the deprecation period was over or whatever other text you wanted there.

mtreinish (approved):
With the exception of one copy paste error inline I think this is fine to merge. There are still some small questions/musings and steps that diverged a bit from what we implemented but that's fine for an rfc as it's a document of a plan not necessarily everything that was implemented. But not worth blocking or changing at this point since the rfc is basically all implemented at this point

PR #46, opened 2023-08-22
Qiskit Ecosystem

frankharkins:


mickahell:
Maybe adding quality filter about on which bases are we accepting a project ? 
Could be one or any of those or anything else :
- they pass automate checks (and so which one)
- they pass manual checks
- ...

PR #47, opened 2023-08-22
Overhaul Qiskit Experiments

nkanazawa1989:
This is the proposal for refactoring plan of Qiskit Experiments. See details for the main document.

Since Qiskit Experiments project is moved to Qiskit-Extensions, where we can rely more on the community, this document includes a long educatory introduction for active developers who newly join this project.


0000-overhaul-qiskit-experiments.md
```diff
@@ -0,0 +1,638 @@
...
+
+## Summary
+
+[Qiskit Experiments](https://github.com/Qiskit-Extensions/qiskit-experiments) (QE) is a qiskit extension project that provides a basic framework 
+for calibration and characterization experiments together with the built-in experiment library.
+The ultimate goal of this project is the enablement of maintenance workflow for a utility-scale quantum processors, which may consist of 
+few hundreds or maybe thousands of qubits.
+This document spots current performance bottlenecks and describes how to overcome the performance issues.
+
+As described in the [project documentation](https://qiskit.org/ecosystem/experiments/tutorials/intro.html#what-is-qiskit-experiments), QE consists of
```
eliarbel, line 19: It seem that this RFC handles 2 big proposed changes: 

1. Improving analysis (hence QE) scalability 
2. Breaking `ExperimentData` to more manageable pieces (via executors). 

Obviously 2 facilitates 1 but we can focus on 1 first without going through the steps suggested for 2, if it creates a reasonable overhead with implementing 1 without 2 and then transitioning to 2. So I wonder if this is the case or if doing 1 without 2 is just too much work and these changes should come together

0000-overhaul-qiskit-experiments.md
```diff
@@ -0,0 +1,638 @@
...
+
+I also scrutinized the performance of the current framework with a statistical profiler.
+Since the percentage of function detection varied significantly from run to run, I only show qualitative results here.
+I found following subroutines tend to frequently show up in the frame graph.
+
+#### Initialization of `ExperimentData` sub-container
+
+Because composite analysis needs to instantiate new data container to run component analyses,
+calling the constructor of this expensive class consumes significant time in heavily batched experiment.
+Especially, acquisition of a service object and qiskit version information is the expensive part.
```
eliarbel, line 98: Should we recommend working in a Python prompt for some use-cases, such that this overhead is paid once?

0000-overhaul-qiskit-experiments.md
```diff
@@ -0,0 +1,638 @@
...
+
+## User Benefit
+
+A user or system agent who heavily rely on the parallel or batch execution feature will dramatically benefit from this overhaul. 
+These users typically be a responsible for the system deployment or maybe a researcher who benchmarks a novel gate with larger application circuits. 
+This kind of work requires users to run a chain of calibration and monitoring experiments at a scale of the quantum processor.
+If experiments are conducted in the dedicated queue mode or with a [runtime in a session](https://qiskit.org/ecosystem/ibm-runtime/how_to/run_session.html), 
+the significant overhead in the local analysis task may also deteriorate the machine efficiency of the remote quantum computing system.
+
+With the proposed overhaul plan, we aim to achieve the T1 analysis wall time of < 5 sec at nq=100 either with a high-end laptop or workstation (sorry we don't have proper benchmark suite for QE!).
```
eliarbel, line 132: I've started something here: https://github.com/eliarbel/qiskit-experiments-benchmarks. We should discuss if and how we want to leverage this or something like this for formalizing a benchmark framework for QE

0000-overhaul-qiskit-experiments.md
```diff
@@ -0,0 +1,638 @@
...
+At a scale of 1000 qubit experiment, maybe human experimentalist doesn't have a chance to visually confirm every experiment result.
+We could conditionally create a figure for unexpected experiment result (e.g. very poor chi-squared value, too short T1 value for T1 analysis) 
+to help a human experimentalist to scrutinize the device issue. Otherwise, we could create the result figure on the fly when it is accessed.
+
+### Proposal of new workflow
+
+Following sequence diagram proposes new workflow of QE run. 
+Note that now we have four local components, `Experiment`, `Analysis`, `Executor`, and `ExperimentData`.
+The experiment data becomes a data container which also implements the interface to the experiment service (this feature can be mix-in).
+Executor has two sub-executors `JobExecutor` and `AnalysisExecutor`.
```
eliarbel, line 250: Potential confusion with the backend's JobExecutor, thus I'd rename it to `ExperimentExecutor`, or `ExperimentRunner` if it comes to be a 1st-class citizen type in QE.

0000-overhaul-qiskit-experiments.md
```diff
@@ -0,0 +1,638 @@
...
+
+```python
+exp_data = MyExperiment(
+    (0,), backend=backend, **options
+).run().block_for_results()
+```
+
+If we have enough bandwidth to support end-user migration, alternative implementation is worth considering.
+
+
```
eliarbel, line 587: I think we need more discussion around the alternative approach (i.e. getting rid of `BaseExperiment.run`), which indeed brings us closer to the runtime way of running. I don't know how acute is the single line of execution capability to our users. Personally I prefer 3 vertical calls instead of 3 horizontal (chained) calls; this way the code better captures the relevant steps, it's easier to debug if needed, profile etc.. But in any case we can keep `BaseExperiment.run` for compatibility (even after the deprecation period) in parallel of having executor-base flow supported.

0000-overhaul-qiskit-experiments.md
nkanazawa1989: Yes, you are correct. 1 and 2 can be done in parallel, and currently @to24toro is investigating the JAX fitter. We can also improve figure generation. However seems like refactoring of composite analysis is very important to improve the user experience. Since composite analysis is tied to the experiment data, so other developer can push forward 2.

0000-overhaul-qiskit-experiments.md
nkanazawa1989: Can you elaborate? I am thinking to introduce simpler container that doesn't have connection to experiment service. I think current design of `_run_analysis(self, experiment_data)` gives too much freedom to experiment developers, since technically they can modify remote data during analysis. This could introduce some edge case that we don't want to support.

0000-overhaul-qiskit-experiments.md
nkanazawa1989: I don't have access to the repo, but I remember you showed a handmade benchmark suite. If we want to continuously monitor the performance, I guess we need to setup some bare metal server. @mtreinish knows how to setup the benchmak.

0000-overhaul-qiskit-experiments.md
nkanazawa1989: That's good point. I replaced it with `CircuitExecutor`, which aligns with circuit-runner runtime working behind the scene. Note that `ExperimentRunner` is `Executor` itself, because experiment = job + analysis execution.

[3382f99](https://github.com/Qiskit/RFCs/pull/47/commits/3382f9995293fc30063b3fbe2b1edb3b4edad797)

0000-overhaul-qiskit-experiments.md
nkanazawa1989: If I can design the pattern from scratch without considering the backward compatibility, I would propose something like this.

```python
with Executor(backend, shots=1000, block=True) as exp_data:
    qe.run(MyExperiment((0,), **options))  # get executor object from the contextvar
```

Context is very powerful, for example, I don't need explicit composite experiment classes:

```python
with Executor(backend, shots=1000, block=True) as exp_data:
    with qe.parallel():
        for q in qubits:
            qe.run(MyExperiment((q,), **options))  # get executor object from the contextvar
```

0000-overhaul-qiskit-experiments.md
eliarbel: I was referring to caching service object and qiskit version in the a python session (say IPython). A container without experiments service is obviously useful for certain use-cases, but what about those use-cases that a user will require results DB access?

0000-overhaul-qiskit-experiments.md
eliarbel: That's nice but I would also consider a non-context based syntax, just not to enforce using `with` blocks for users who prefer the standard code layout for some reason

0000-overhaul-qiskit-experiments.md
eliarbel: Ideally we should have a bare metal server and something like ASV. Let's have a separate discussion about this aspect.

0000-overhaul-qiskit-experiments.md
nkanazawa1989: I meant analysis usually doesn't require results DB access, so we can introduce simplified internal data container that doesn't have experiment service. Caching service object is easy and effective, but do we really need service object to run experiment analysis?

0000-overhaul-qiskit-experiments.md
```diff
@@ -0,0 +1,638 @@
...
+## Questions
+How to communicate with community developers?
+How to avoid breaking API change?
+What is the reasonable timeline for migration?
+
+Another important situation that we may need to cover is the break of the executor run.
+Job queueing in a quantum system may take very long time (sometime more than a day) if you don't reserve the device, 
+and we often terminate the kernel where the circuit executor is actively running.
+In this situation, we may want to restart from the analysis, once after all jobs successfully complete.
```
coruscating, line 596: I think we should also add an option to not run the analysis automatically after job completion, like @TsafrirA's use case.

0000-overhaul-qiskit-experiments.md
```diff
@@ -0,0 +1,638 @@
...
+
+```
+
+It's noteworthy that one can also run experiment with transpiled circuits by directly using executor interface.
+See [Qiskit-Experiments/#1222](https://github.com/Qiskit-Extensions/qiskit-experiments/pull/1222) for background.
+
+```python
+exc = Executor()
+exc.circuit_executor.backend = some_backend
+exc.circuit_executor.add_circuits(transpiled_circuits)
```
coruscating, line 534: I like this usage pattern. Where would the circuit validation hook be?

0000-overhaul-qiskit-experiments.md
coruscating: If runtime sessions will become the default for most users, I think the context based syntax as the Q-E default is fine. I like the idea of keeping `BaseExperiment.run` for compatibility but promoting this context-based syntax as the standard pattern.

coruscating (commented):
This is really nice work!

0000-overhaul-qiskit-experiments.md
nkanazawa1989: I added more comment about extendability in [a99c946](https://github.com/Qiskit/RFCs/pull/47/commits/a99c94639c13fdbc853c09bd77bca0e3668318dc). Let's separately discuss execution pattern and focus on internal rework in this RFC. We need more user feedback to decide the interface.

0000-overhaul-qiskit-experiments.md
nkanazawa1989: Yes. We can continue to support `Experiment.run(analysis=None)`. In this case the analysis executor will receive an empty task dependency graph and run nothing.

0000-overhaul-qiskit-experiments.md
nkanazawa1989: Could you please elaborate? This is something to validate the circuit metadata against analysis class?

coruscating: Yes, I was thinking something to validate that the metadata and circuit output formats are what's expected by the analysis class. We don't necessarily have to implement it for every analysis class, but having a template hook would be useful.

nkanazawa1989: Added description in [b83b848](https://github.com/Qiskit/RFCs/pull/47/commits/b83b8483febb424c1ba53267fcf6c501f89255f7). Validation for composite experiment circuit might become too complicated to implement. Probably we need to simplify the metadata structure.

coruscating (approved):
LGTM! 🚀

PR #50, opened 2023-09-13
Variable-declaration and storage representation in Qiskit

jakelishman:
This is a proposal and design comments on how to begin adding the representations of declared classical variables and stores to Qiskit, including input variables, which correspond to the `input` keyword in OpenQASM 3, and can be used to define parametric input of arbitrary types into Qiskit.

For now, this RFC does not propose expanding Qiskit's classical type system, so only `Bool` and `Uint` remain valid, and these parameters cannot yet be used in gate-argument positions.  This is still strongly desired, just not for this RFC.

---

Apologies if there's any thoughts that trail off, or places that need expansion - I needed to push this now so people can start to read it, rather than keeping it all local for longer.  It's also a bit wordier and less organised than I'd like because I haven't really gone back through and edited it.

### DIFF
--- a/0000-classical-stores.md
+++ b/0000-classical-stores.md
@@ -1,0 +1,493 @@
+# Classical storage and load instructions
+
+| **Status**        | **Proposed/Accepted/Deprecated** |
+|:------------------|:---------------------------------------------|
+| **RFC #**         | ####                                         |
+| **Authors**       | Jake Lishman (jake.lishman@ibm.com)          |
+| **Submitted**     | 2023-09-07                                   |
+| **Updated**       |                                              |
+
+
+## Summary
+
+In scope:
+
+- Add new, block-scoped classical `Var` storage locations to `QuantumCircuit` that will (long term) replace `ClassicalRegister` and `Clbit`.
+- Introduce instructions to manually assign `Expr` nodes to a storage location.
+- Allow input and output from these storage locations at the start/end of program execution.
+
+Out of scope:
+
+This RFC does not propose making `Expr` nodes available in any more places than they already are, nor adding any new types to the system.
+In particular, this RFC does not yet address the question of making `Expr` nodes usable in `params` locations (such as the angle inputs to gates).
+
+
+## Motivation and Benefit
+
+Qiskit needs a way to represent the sorts of "parametric update" that we expect hardware vendors to be implementing in the near future.
+For example, we will ideally be able to use a single compiled circuit (both in the Qiskit and hardware senses of the term "compiled") that accepts certain parameters as inputs at runtime, which may affect conditional expressions or gate angles.
+This RFC only addresses the question of defining new classical-variable storage locations and writing to them for Qiskit within its existing typed `Expr` system, which is currently limited to unsigned integers and Booleans and cannot be used for gate angles.
+
+For example, Pauli-twirling workloads want to do something like
+```python
+from qiskit.circuit import QuantumCircuit
+from qiskit.circuit.classical import expr
+
+qc = QuantumCircuit(2)
+with qc.if_test(expr.value(TWIRL_BITS_0)):
+    qc.x(0)
+with qc.if_test(expr.value(TWIRL_BITS_1)):
+    qc.x(1)
+```
+where `TWIRL_BITS_*` are some values defined by the particular set of twirls to be done, and are known before the circuit begins execution, but not _necessarily_ at the time of the definition of the circuit, or the circuit should be run with many different values for them.
+These may also want to be compared to the results of mid-circuit measurements, or have arbitrary run-time classical expressions built from them.
+Right now, Qiskit provides no way to parametrise `if` conditions in any form; even the compile-time rewriting of `assign_parameters` will not support them, because `ParameterExpression` objects are (deliberately) not valid in `condition` fields.
+
+Similarly, a user may need to calculate a value based on some measured bits, potentially dependent on some `if` condition, and store the value in classical memory for later use in the circuit.
+In other words, they need variable assignment at runtime.
+
+
+## Design Proposal
+
+At a high level, this proposal concerns adding concepts equivalent to the OpenQASM 3 `input` (and potentially `output`) and variable-assignment-by-expression statements to Qiskit, within the classical types we already support.
+I do not yet propose removing `Clbit` or `ClassicalRegister`, nor trying to change them into some other form; for the time being, they retain their "special-case" status.
+
+The additional user interface for defining and working with variables on `QuantumCircuit` is:
+
+```python
+class QuantumCircuit:
+    def __init__(
+        self,
+        ...,  # existing structure
+        *,
+        inputs: Iterable[expr.Var] = (),
+        captures: Iterable[expr.Var] = (),
+        declarations: Mapping[expr.Var, expr.Expr] | Iterable[Tuple[expr.Var, expr.Expr]] = (),
+    ): ...
+
+    def add_var(self, name_or_var: str | expr.Var, / initial: expr.Expr) -> expr.Var: ...
+    def add_capture(self, *vars: expr.Var): ...
+    @overload
+    def add_input(self, name_or_var: str, type: types.Type, /) -> expr.Var: ...
+    @overload
+    def add_input(self, name_or_var: expr.Var, /) -> expr.Var: ...
+
+    def store(self, location: expr.Expr, value: expr.Expr, /): ...
+
+    @overload
+    def get_var[T](self, name: str, *, default=T) -> Union[expr.Var, T]: ...
+    @overload
+    def get_var(self, name: str, *, default=NotGiven) -> expr.Var: ...
+    def has_var(self, var: Union[str | expr.Var]) -> bool: ...
+
+    # All variables, including inputs and captures.
+    def iter_vars(self) -> Iterable[expr.Var]: ...
+
+    # All do what they say on the tin.
+    def iter_declared_vars(self) -> Iterable[expr.Var]: ...
+    def iter_input_vars(self) -> Iterable[expr.Var]: ...
+    def iter_captured_vars(self) -> Iterable[expr.Var]: ...
+```
+
+`QuantumCircuit.add_var` and `.add_input` are for declaring a variable with an owned storage location, the difference being that `add_var` is an internal variable and requires an initialiser, while `add_input` marks this as in input variable to the circuit and cannot have an initialiser.
+The variable remains in scope for the whole body of the circuit.
+Variables are not automatically inherited by nested scopes; they must be explicitly closed over when building an inner scoped `QuantumCircuit` instance using `QuantumCircuit.add_capture`.
+Only `add_input` takes a `Type`, and even then, only if the input is given as a `str` name.
+All other forms read the type from the `expr.Var` or the initialiser `expr.Expr`.
+It is an error if the initialiser expression is not of a type that implicitly casts to the type of the `Var`.
+If a `str` name is given in `add_var`, the type of the resulting expression will be exactly the type of the initialiser, and this should be manually cast to the desired type if it is not the exact required type.
+
+The initialiser of `QuantumCircuit` is expanded to allow setting these as well.
+This is not intended to be how the _first_ circuit a user creates constructs its variables in general; it is meant to make it easy to build more circuits similar to the first circuit by reusing the same `Var` instances that are already held.
+In the initialiser, all the variables must be given as `expr.Var`, since if a string was permitted, a user would then have to do `get_var` to retrieve the `Var` before they could do anything anyway, at which point they can use `add_var` which is just as ergonomic.
+The keyword `declarations` must take some form of mapping from the `expr.Var` values to the initial expressions.
+A user must supply a mapping or iterable that will yield the `Var` nodes in a suitable order for declaration; it will be an error if an initialiser expression relies on a `Var` that has not yet been seen.
+Python's built-in `dict` yields its items in insertion order (up to removals), so this is not expected to be an onerous ask for the API.
+
+`QuantumCircuit.store` inserts a `Store` instruction, which at QPU runtime evaluates `value` and stores it in the memory location that `location` resolves to.
+The `location` must evaluate to a writeable memory location.
+In this initial RFC this is limited to `Var` instances, though in the future one can imagine some `expr.Index` node or the like existing, for runtime indexing into objects that are array-like in bits.
+
+`QuantumCircuit.get_var` and `.has_var` are convenience methods for the user.
+`.has_var` is a simple Boolean checker akin to `dict.__contains__`.
+`.get_var` is akin to `getattr`; if the default is not given it raise an error, while if a default is given, that is returned instead.
+This design is to make it ergonomic both when all variables are expected to be present (one need not check manually that an actual `Var` was returned; it's sufficient to _not_ catch the exception), and when an action should be taken _only_ if the variable is present, such as in
+```python
+if (var := qc.get_var("a", None)) is not None:
+    # Do something with `var`, which is guaranteed to be a `Var`,
+    # but no extra `try`/`except` is needed if it might not exist.
+```
+These methods are present to solve a common user paper-cut when working with `Parameter` values, where they can't retrieve the parameter only by its string name despite circuits enforcing that these instances have unique names or are referentially identical within a circuit.
+
+The iterator methods `iter_vars`, `iter_declared_vars`, `iter_input_vars` and `iter_captured_vars` all return single-use iterator objects over the relevant variables.
+This choice lets `QuantumCircuit` use a mutable type internally for performance when adding new variables, while allowing iteration through the variables without the necessity of copying the collection when outputting it to the user to avoid them messing with it.
+
+
+### Examples
+
+The example given above in the motivation section would look like:
+
+```python
+from qiskit.circuit import QuantumCircuit
+from qiskit.circuit.classical import expr, types
+
+qc = QuantumCircuit(2)
+twirl_bits_0 = qc.add_input("twirl_bits_0", types.Bool())
+twirl_bits_1 = qc.add_input("twirl_bits_1", types.Bool())
+with qc.if_test(twirl_bits_0):
+    qc.x(0)
+with qc.if_test(twirl_bits_1):
+    qc.x(1)
+```
+
+The result of `qasm3.dumps()` on this circuit will be (subject to header-file discussions, etc):
+```qasm3
+OPENQASM 3.0;
+
+include 'stdgates.inc';
+
+input bool twirl_bits_0;
+input bool twirl_bits_1;
+
+qubit _qubit_0;
+qubit _qubit_1;
+
+if (twirl_bits_0) {
+    x _qubit_0;
+}
+if (twirl_bits_1) {
+    x _qubit_1;
+}
+```
+
+---
+
+Using the control-flow builders will allow scoped declarations, and using a variable declared in an outer builder scope will automatically cause it to become captured within the output.
+For example:
+
+```python
+from qiskit.circuit import QuantumCircuit, QuantumRegister, ClassicalRegister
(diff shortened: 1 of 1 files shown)


0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+Until it's clear how _all_ instructions should be parametrised, I elected to keep the new object in the same sort of special case as the other classical operations (the control-flow operations).
+
+
+### What does the API for `QuantumCircuit.compose` expand to?
+
+`QuantumCirucit.compose` is in effect an inlining operation.
+It's not clear to me whether it should be possible for a circuit to involve variables whose names conflict in scope with the program it's being inlined into, or what if any automatic variable-renaming methods we should make available.
+
+If there are no conflicts in the variable names, it seems expected to me that the lists of input variables would catenate, and so on.
+Similarly, two declared `Var` instances that conflict (either by having incompatible initialisers, or one being an input, one not, etc) cannot be allowed to coexist, but does this mean that the composition should be rejected, or should we implement a renaming?
```
1ucian0, line 431: I think how to handle namespace conflicts should be parametrizable to the `compose` method. It might cover a valid usecase glue circuits referring to the save variable.

jakelishman: The only time in this system that you could have two circuits really validly referring to the same variable without a redefinition via the initialiser is in the cases:
- base circuit declares the `Var` as an `input`, inlined declares it as input or closure
- base circuit declares the `Var` as a closure, inlined declares it as closure
- base circuit initialises the `Var`, inlined declares it as closure

Those are all largely ok, but given the nature of how closures work, you have to have a reference to the `Var` already, so I don't think there's much problem in `compose`; we can easily enforce that it only works with the same `Var`.

The issues are if I've got two circuits like
```python
base = QuantumCircuit()
a1 = base.add_var("a", types.Bool(), expr.value(True))

inlined = QuantumCircuit(
a2 = inlined.add_var("a", types.Bool(), expr.value(False))
```
now there are two variables named "a" that are not the same, and naive inlining re-using the same named storage location would cause `a2` to shadow `a1`, and for the ensuing data model to work, we'd already have to rewrite all expressions in `inlined` to use `a1` instead of `a2`.

My preference here would likely be to reject compositions with any variable clashes that would require complete variable rewriting, just for ease in the initial implementation; I think there's nuances to classical-function inlining that I'm not super confident of at the moment to allow something we're not 100% of the utility of.

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+Currently this RFC does not propose this to avoid scope creep, but I don't know if it's something that ought to be done now.
+
+
+## Future Extensions
+
+### Add support for circuit outputs
+
+OpenQASM 3 supports an `output` specifier on variables for rich returns from programs.
+Qiskit currently doesn't have a way of representing this through its result types, and it's not clear when hardware vendors will have support for this.
+For ease of user expectations, it's left til later.
```
1ucian0, line 466: Maybe I'm bending things too much. I feel the current approach is that every classical wire has an implicit `output` right now.

jakelishman: They do, and the OQ3 spec even defines that behaviour to make it consistent-ish with OQ2-era Qiskit behaviour.  In practice, I shied away from defining this because I think the harder questions to answer are hardware-vendor side.

0000-classical-stores.md
kevinhartman, line 431: IMO, names defined explicitly by the user with conflicts should be an error, by default. But perhaps it's worth providing a mechanism to override this behavior. I'd think a new boolean parameter on `compose` would be sufficient for this, e.g. `shared_vars: Bool`. A value of `True` meaning that rather than erroring-out on conflict, conflicting vars should be checked for compatibility and shared. Then, it's up to the user to first rename any conflicting vars that aren't intended to be the same prior to composition.

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+
+Finally, the potential mental load from having two separate methods is mitigated because we do not expect users to manually call `add_capture` themselves.
+The control-flow builders can intercept all `expr.Expr` values coming in, and automatically cause the scope under construction to close over any undefined variables, if they exist in the outer circuit.
+`QuantumCircuit.add_capture` would only be necessary when manually constructing a `QuantumCircuit` scoped block separately, where the user must manage all the context themselves.
+This is behaviour libraries may choose to do, but we already _strongly_ suggest that users use the control-flow builder interface.
+
+
+#### Allow `str` names in `expr` builder constructions instead of using `QuantumCircuit.get_var` to get there
+
+It feels tempting to allow things like `expr.bit_and("a", 7)` to automatically promote `a` to a `Var`.
```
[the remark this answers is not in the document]
jakelishman: Can you give any examples of what you mean? I'm having a hard time seeing an API call structure that would be sensible here.

0000-classical-stores.md
1ucian0: I would prefer the Kevin's `compose`-parametrisation approach.

jakelishman: Yeah, this sounds totally reasonable to me. If there's no objections, for an initial implementation I suggest we just put in "error on conflict", and add any additional flags to implement a partial sharing once we find a compelling use case.

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+#### Give `Var` nodes backing `Clbit`s
+
+I seriously considered this to the degree of holding up publishing this RFC to pivot to this design.
+It has the benefit of:
+- making `ClassicalRegister` a little less special-cased; it's a variable with contained bits.
+- it would be possible to represent a `measure` directly into a bit of a new-style variable
+
+However, in the end, I have many reasons that I think this would not the correct decision:
+
+- When
```
1ucian0, line 342: Should this bullet be removed?

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+
+### Alternatives to smaller details
+
+#### Automatic closure over variables declared as `input`
+
+In the current system, it is not valid to have a circuit that has both closure variables _and_ input or output variables.
+Instead of having separate `add_input()` and `add_capture()` methods, one could imagine that inner scopes could `add_var` variables to be either of type "input" or a new type "closure".
+The reason to choose to have a separate method for these semantics is that the signature of the `add_var` method would become quite unpleasant if it needed to do all these things; it already has multiple signatures because it needs to take either `(var,)` or `(name, type)`.
+
+It is not possible to close over a variable with just its name and type, because we'd also need the information about which root allocation it refers to to build up the `DAGCircuit` data-flow graph, and we don't gain that information until the circuit is embedded within another.
```
1ucian0, line 375:
```suggestion
It is not possible to close over a variable with just its name and type, because we'd also need the information about which root allocation it refers to build up the `DAGCircuit` data-flow graph, and we don't gain that information until the circuit is embedded within another.
```

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+`QuantumCircuit.get_var` solves the latter real-world problem, without introducing awkwardnesses into the type signatures of other functions.
+
+
+#### Allow `Measure` to write into `Expr` l-values like `Store`
+
+This RFC currently has a distinction between `Measure` which still uses the old `cargs` form and consequently writes into a `Clbit` instance, and `Store` that contains the memory locations inherently, and thus must be extracted by things like `DAGCircuit.apply_operation_back` (and would be easy to miss).
+The main reason not to expand `Measure` directly in this RFC is the complexity; `Measure` is an _exceptionally_ common operation (perhaps the only one that is almost guaranteed to be in a complete quantum program), and it's not yet clear what the full expansion path here should be.
+Allowing a direct measurement into an `Expr` would involve updating every consumer of `CircuitInstruction.clbits` to know that the values may not always be `Clbit` instances any more, which would be too breaking a change.
```
1ucian0, line 402:
```suggestion
Allowing a direct measurement into an `Expr` would involve updating every consumer of `CircuitInstruction.clbits` to know that the values may not always be `Clbit` instances anymore, which would be too breaking a change.
```

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+The additional user interface for defining and working with variables on `QuantumCircuit` is:
+
+```python
+class QuantumCircuit:
+    def __init__(
+        self,
+        ...,  # existing structure
+        *,
+        inputs: Iterable[expr.Var] = (),
+        captures: Iterable[expr.Var] = (),
```
blakejohnson, line 64: Are you at all concerned about the conflict with the other way "capture" is used in openpulse? https://openqasm.com/language/openpulse.html#capture-instruction

blakejohnson: I see below, though, that you are using this exactly in the sense of C++ "captures" for closures. Well, it is an unfortunate word collision.

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+
+The initialiser of `QuantumCircuit` is expanded to allow setting these as well.
+This is not intended to be how the _first_ circuit a user creates constructs its variables in general; it is meant to make it easy to build more circuits similar to the first circuit by reusing the same `Var` instances that are already held.
+In the initialiser, all the variables must be given as `expr.Var`, since if a string was permitted, a user would then have to do `get_var` to retrieve the `Var` before they could do anything anyway, at which point they can use `add_var` which is just as ergonomic.
+The keyword `declarations` must take some form of mapping from the `expr.Var` values to the initial expressions.
+A user must supply a mapping or iterable that will yield the `Var` nodes in a suitable order for declaration; it will be an error if an initialiser expression relies on a `Var` that has not yet been seen.
+Python's built-in `dict` yields its items in insertion order (up to removals), so this is not expected to be an onerous ask for the API.
```
blakejohnson, line 105: Does python guarantee this for `dict` or is it just an artifact of the current implementation?

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+    def __init__(
+        self,
+        ...,  # existing structure
+        *,
+        inputs: Iterable[expr.Var] = (),
+        captures: Iterable[expr.Var] = (),
+        declarations: Mapping[expr.Var, expr.Expr] | Iterable[Tuple[expr.Var, expr.Expr]] = (),
+    ): ...
+
+    def add_var(self, name_or_var: str | expr.Var, / initial: expr.Expr) -> expr.Var: ...
```
blakejohnson, line 68: I think you are missing a `type: types.Type` between `name_or_var` and `/`.
```suggestion
    def add_var(self, name_or_var: str | expr.Var, type: types.Type, / initial: expr.Expr) -> expr.Var: ...
```

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+For example:
+
+```python
+from qiskit.circuit import QuantumCircuit, QuantumRegister, ClassicalRegister
+from qiskit.circuit.classical import expr, types
+
+qr = QuantumRegister(3)
+cr = ClassicalRegister(3)
+qc = QuantumCircuit(qr, cr)
+mask = qc.add_var("mask", types.Uint(3), expr.value(5))
```
blakejohnson, line 175: I find this a little hard to scan as an assignment `uint[3] mask = 5`, but I am struggling to come up with a good alternative. I guess it does read a little better when providing the `initial=` keyword.

0000-classical-stores.md
jakelishman: It used to be a CPython implementation detail, but since 3.7 it's been a part of the language spec.

0000-classical-stores.md
jakelishman: No, it's deliberately not there because it's always inferable, so having it would only really offer an opportunity for desynchronisation and errors.

0000-classical-stores.md
jakelishman: Yeah, it's not perfect for sure. This RFC is principally meant to be the low-level methods and implementation - I think a higher order interface would be a follow up.

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+We currently have a problem in the control-flow operations where the exact classical registers must be used in inner blocks (if they are to be accessed), and this constrains the data in the `Clbit` binding between the `CircuitInstruction` and the individual blocks' `.clbits` in a way that is not clear at all, and introduces many possibilities for data incoherence.
+Requiring closures to be explicitly stated sidesteps all of this; there is no longer a "binding" step, and the validity of all the operations can be enforced during construction only.
+Holding a `QuantumCircuit` then contains all the information needed to work with the new objects within it.
+
+Of existing circuit methods:
+- `QuantumCircuit.copy_empty_like` would need to expand to include variables in the copy.
+- `QuantumCircuit.from_instructions` needs to gain the same keyword arguments as the `__init__`.
+- `QuantumCircuit.compose` and `QuantumCircuit.tensor` certainly need to have _some_ handling for this (even if it's just erroring if both circuits contain variables), but it's not clear to me how variable collisions should be handled.
```
blakejohnson, line 271: I don't think you want to just error if both circuits contain variables. I think the right behavior depends on whether we think of `compose` as introducing a new scope or as inlining one circuit into another.

0000-classical-stores.md
jakelishman: More precisely, a "closure" is a structure (the function plus an environment) that may have "captures" (the variables from the environment), though we often are a bit lax in terminology and use "closure" to refer to both. I was trying to use the proper name here.

0000-classical-stores.md
blakejohnson: In all your examples below you provide a type, though, e.g.
```
qc.add_var("mask", types.Uint(3), expr.value(5))
```

0000-classical-stores.md
```diff
@@ -0,0 +1,493 @@
...
+- it would be possible to represent a `measure` directly into a bit of a new-style variable
+
+However, in the end, I have many reasons that I think this would not the correct decision:
+
+- When 
+- There's 5 pointers of storage space required per Python-space `Clbit` that's just wrapped by a `Var`, not to mention the additional costs of tracking the bit within `QuantumCircuit`.
+  For a `double`, that's over 2kB of extra tracking, even if we do nothing with it.
+- It implies that we need to track each of these `Clbit` wires separately in the DAG, even though the data-flow is bundled - again, lots of unnecessary memory use.
+- There is a natural split in hardware between a `Clbit`, which is a target for a `measure` instruction and is likely written to in controllers very close to the QPU hardware, and classical variables that can be in outer control systems.
+  For example, gathering the result of a `measure` into the particular bit of a `uint[8]` might involve more expensive transfer and synchronisation operations than adding together two `uint[8]` objects.
```
blakejohnson, line 347: This distinction makes me a bit uncomfortable, particularly because when you serialize a `Var` of type `bit` or a `Clbit` to OpenQASM they should end up indistinguishable.

0000-classical-stores.md
blakejohnson: This actually touches on my primary question about this RFC, which was that I was expecting it to simultaneously address circuit outputs. At the very least, I can imagine wanting to NOT return some values that I used as intermediaries in my circuit.

blakejohnson (commented):
Very strong proposal, @jakelishman. My two primary pieces of feedback inline. But, to pull them out here for visibility they are:
1. With this change, `ClassicalRegister` now seems redundant with  bit-typed `Var`s. Can these be brought together?
2. What about declaring variables as `output`s? It is oddly asymmetrical to introduce support for `input`s and not also introduce `output`s.

0000-classical-stores.md
blakejohnson: I see that you take the "inlining" view below, which feels right to me. I am good with the approach you agreed to with @kevinhartman.

0000-classical-stores.md
jakelishman: I can certainly imagine that too.  My intent with this RFC wasn't to imply that all variables will always be `output` for sure.  I'm fine if we'd like to discuss including `output` in this RFC directly (especially now it's pushed to 1.0) - my initial concern was that I didn't want to specify an interface for something that (at least) IBM hardware definitely has no way of representing in the timescales that Qiskit could implement it.

I think it's relatively easy to imagine ways to designate `output` variables in this form. Some (non-exclusive) options are
- have an `output=True` keyword on `add_var` (but not `add_capture` or `add_input`)
- have an `add_output` function with the same signature as `add_var`
- have an `outputs=...` field in the `QuantumCircuit` constructor

The bulk of the remaining `output` work will then be in the results interfaces of the primitives and `backend.run`, and I think the design process might be smoother and more focussed for other affected teams if we separate that out.

0000-classical-stores.md
jakelishman: Yes - I _knew_ there was somewhere I'd left an unfinished thought as I was pushing it, but I just couldn't spot it!

0000-classical-stores.md
blakejohnson: All of those options seem reasonable.

0000-classical-stores.md
jakelishman: Fixed in 1dc9c4569f9d9274246d2baae98eb2961cf467bb.

0000-classical-stores.md
jakelishman: Fixed in 65f9fa2d33e8405186f92e54b5cb25a53beb6e17.

jakelishman:
I've (finally) updated this RFC to respond to the comments above, thanks.

For visibility on Blake's comment above:

> 1. With this change, `ClassicalRegister` now seems redundant with bit-typed `Var`s. Can these be brought together?
> 2. What about declaring variables as outputs? It is oddly asymmetrical to introduce support for inputs and not also introduce outputs.

1. See discussion in https://github.com/Qiskit/RFCs/pull/50#discussion_r1378129090 and newly in the RFC, but the gist is that we absolutely do want to bring them together, just there's technical and stack reasons to hold off from doing it within this RFC.
2. See https://github.com/Qiskit/RFCs/pull/50#discussion_r1383529454.

1ucian0:
@blakejohnson and @kevinhartman ,

if you are happy with the way your questions were addressed and you documented all your opinions on the matter, do you mind approving this PR? I will merge it after.

0000-classical-stores.md
```diff
@@ -0,0 +1,516 @@
...
+- `ClassicalRegister` would _still_ be special cased, however, because it does not fundamentally "own" its bits; it's possible to have aliased bits in more than one register.
+  This is very fiddly to work with in general, is highly unlikely to have any real hardware support in the near term, and it's unclear how it's useful.
+
+#### Give `Store` first-class support as an alternative to `CircuitInstruction`
+
+The RFC currently proposes to jam `Store` into being a subclass of `Instruction` that has inherent classical bits (similar to `Instruction.condition` or the `ControlFlowOp` subclasses).
+This means that all consumers of `QuantumCircuit` that care about data-flow orderings and display must know to check for this particular subclass, and include custom code to handle this.
+Similarly, the `DAGCircuit` would need to grow an alternative to `DAGOpNode` that could represent this with new first-class support.
+
+A strong alternative would be to have different versions of the `CircuitInstruction`/`DAGOpNode` circuit context object that better represent the "special" classical operations such as `Measure`, `Store` and the `ControlFlowOp` subclasses.
```
kevinhartman, line 360: Worth noting that this would imply that `CircuitData` introduced in https://github.com/Qiskit/qiskit/pull/10827 would need to gain awareness of these special classical operations (i.e. via separate methods like `append_store` or with subclass instance checking through its `append`, in addition to internal tracking in either case).

It's certainly doable, but it's useful to know that `CircuitData` (though currently private) will likely need more awareness/introspection of the instructions it stores, esp. as we oxidize more parts of `QuantumCircuit`.

kevinhartman (approved):
Looks good to me.

This seems like the right size of step to take as well, and doesn't lock us into anything that isn't absolutely necessary now.

blakejohnson (approved):

PR #51, opened 2023-10-04
Updated Estimator interface

ihincks:


0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+
+```python
+# construct two circuits, the first with 2039 Parameters, the second with no Parameters
+circuit1 = QuantumCircuit(9)
+...
+circuit2 = QuantumCircuit(2)
+...
+
+# specify 128 different parameter value sets for circuit1, in a 4x32 shape
+parameter_values1 = np.random.uniform((4, 32, 2039))
```
garrison, line 74: I don't think that's what this line does.
```python
>>> np.random.uniform((4, 32, 2039))
array([  2.26565035,   8.87576619, 446.18998942])
```

ihincks: Thanks, fixed in e400b2484de6a.

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+            # Coerce into old form
+            if parameter_values is None:
+                parameter_values = itertools.repeat(None)
+            if isinstance(circuit, QuantumCircuit):
+                circuit = [circuit]
+            if isinstance(observables, (BaseOperator, PauliSumOp, str)):
+                observables = [observables]
+
+            # Coerce old form into `EstimatorTask`s
+            tasks = [EstimatorTask.coerce(task) for task in zip(circuits, observables, parameter_values)]
```
t-imamichi, line 384: It would be good to add a validate step of numbers of circuits, observables, and parameter_values before generating tasks. Otherwise, the number of tasks could be smaller than expected e.g. if the number of observables is smaller than that of circuits.

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+        if tasks is None:
+            # Verify values in old API
+            if not (circuits and observables):
+                raise ValueError(f"Need both circuits and observables for old API, got circuits={circuits}, observables={observables}")
+            
+            # Coerce into old form
+            if parameter_values is None:
+                parameter_values = itertools.repeat(None)
+            if isinstance(circuit, QuantumCircuit):
+                circuit = [circuit]
```
t-imamichi, line 379:
```suggestion
            if isinstance(circuits, QuantumCircuit):
                circuits = [circuits]
```

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+            if isinstance(observables, (BaseOperator, PauliSumOp, str)):
```
t-imamichi, line 380: Since opflow is deprecated and will be removed soon, we might be able to remove it.
```suggestion
            if isinstance(observables, (BaseOperator, str)):
```

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+parameter_values = BindingsArray(np.random.uniform(size=(5,)))
+parameter_values.shape == (5,)
+observables = ObservablesArray([[Pauli("III")], [Pauli("XXX")], [Pauli("YYY")], [Pauli("ZZZ")], [Pauli("XYZ")]])
+observables.shape == (5,1)
+>> result_bundle.shape == (5,)
+
+# Outer/Product
+parameter_values = BindingsArray(np.random.uniform(size=(1,6)))
+parameter_values.shape == (1,6)
+observables = ObservablesArray([[Pauli("III"), Pauli("XXX"), Pauli("YYY"), Pauli("ZZZ"), Pauli("XYZ"), Pauli("IIZ")]])
```
1ucian0, line 240: Shouldn't they be 4 observables? 
```suggestion
observables = ObservablesArray([[Pauli("III"), Pauli("XXX"), Pauli("YYY"), Pauli("ZZZ")]])
```

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+# Broadcast single observable
+parameter_values = np.array(1.0)
+parameter_values.shape == ()
+observables = ObservablesArray(["III", "XXX", Pauli("YYY"), "ZZZ", Pauli("XYZ")])
+observables.shape == (5,)
+>> result_bundle.shape == (5,)
+
+# Inner/Zip
+parameter_values = BindingsArray(np.random.uniform(size=(5,)))
+parameter_values.shape == (5,)
```
1ucian0, line 232: consider that `(5,)` does not align with `(5,1)`

0015-estimator-interface.md
1ucian0: the shape (4,1) shouldn't be link this?
```suggestion
observables = ObservablesArray([[Pauli("III")], [Pauli("XXX")], [Pauli("YYY")], [Pauli("ZZZ")]])
```

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+
+With the `Estimator`, it is common for a user to want to estimate many observables of a single circuit. 
+For example, all weight-1 and weight-2 Paulis that are adjacent on the connectivity graph. 
+For a one-hundred qubit device, this corresponds to hundreds of unique estimates to be made for a single circuit, noting that for this particular example, on the heavy-hex graph, in the absence of mitigation, only 9 circuits need to be physically run.
+
+We propose a new `ObservablesArray` object, where each element corresponds to a observable the user wants an estimated expectation value of.
+We propose that the internal data model of this object have an element format
+
+```python
+List[Tuple[str, List[int], float]]
```
blakejohnson, line 196: Are we going to expand the alphabet of allowed characters in a `str` representation from (I, X, Y, Z) to also include (0, 1, +, -, l, r) in order to capture state projectors?

0015-estimator-interface.md
ihincks: Would you be able to write a patch or code suggestion that I can commit?

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+
+where, for example, the observable `0.5 * IIZY + 0.2 * XIII` would be stored as
+
+```python
+[("ZY", [2, 3], 0.5), ("X", [0], 0.2)]
+```
+
+This is proposed instead of `{"IIZY": 0.5, "XIII": 0.2}` anticipating the overhead of storing and working with so many `"I"`'s for large devices.
+
+Note that the above is simply the data model; the `ObservablesArray` object, either in the constructor or special static coercion method, will accept the `quantum_info` objects such as `Pauli` and `SparsePauliOp` that they are used to.
```
blakejohnson, line 207: We will simultaneously drop support for the generic `BaseOperator` given the infeasibility of finding a polynomial-sized representation of operators in the generic case?

0015-estimator-interface.md
blakejohnson: I agree with @1ucian0

0015-estimator-interface.md
ihincks: Yes, I would like to, but my general philosophy with this RFC was to omit features that would be easy to put in a future non-breaking change. If you feel that this does not apply to non-Pauli strings, let me know.

0015-estimator-interface.md
ihincks: (bearing in mind that the main purpose of this block is to convince everyone that there is a feasible migration path, so I'd avoid adding it in a way that introduces clutter)

0015-estimator-interface.md
blakejohnson: Ok, your approach makes sense.

ajavadia:
This is not strictly related to this

ihincks:
> This is not strictly related to this RFC, but it is something that must be changed in the base classes and implemented in the Runtime implementation as well. It's not big enough to merit an RFC on its own, but I'm happy to open an issue if you let me know where.
> 
> For the Estimator output, we are currently returning the mean value, plus a field called 'variance'. I think there's a mismatch in the definition however, because some places it is really the standard error that is returned, not the variance. So we just need to document what this field is and return it correctly. Can we make it the variance, not the standard error? i.e. it should not go to zero with increased number of shots. The user can calculate standard error given the number of shots.

Yes we need to standardize this and make sure we are transparent about what, exactly, we are reporting. I think there are some subtleties when tactics like twirling are involved, and I don't think that multinomial statistics alone are always appropriate in such cases. Similar when drift is significant. But these topics are perhaps too involved to discuss in this thread. @chriseclectic and others will have strong opinions, too.

In terms of the code that is proposed in this RFC, `TaskResult` is admittedly under-developed. @chriseclectic has convinced me that if we are serious about the concept of `BasePrimitive`, then something like `TaskResult` (or subclasses thereof) should be the output of all primitives. But on top of this, we also want to offer some promise about which array-valued outputs are available for each primitive. For the estimator, undoubtedly, at least the EVs. But as you say, some standard form of error bars, too.

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+
+The current `Estimator.run()` method requires that a user provide one circuit for every observable and set of parameter values that they wish to run.
+(This was not always the case via circuit indices, but the history will not be discussed here.)
+It is common, if not typical, for a user to want to estimate many observables corresponding to a single circuit. 
+Likewise, it is common, if not typical, for a user to want to supply multiple parameter value sets for the same circuit.
+This RFC proposes to fix these issues by changing the `Estimator.run()` signature in the following ways: 
+
+ 1. Take the transpose of the current signature; rather than accepting `circuits`, `parameter_values`, and `observables` as three different (iterable) arguments which are to be zipped together, instead localize the distinct tasks to be run via an iterable of triples `(circuit, parameter_values, observables)`.
+ 2. In combination with 1, extend `parameter_values` and `observables` to be array-valued, that is, add the ability to explicity and conveniently specify multiple parameter value sets and observables for a single circuit.
+
```
jyu00, line 28: This RFC only talks about Estimator, but I assume we want to do the same for Sampler, to keep the input format consistent if nothing else?

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+* If necessary, coerce the old-API arguments (now all keyword arguments) into `EstimatorTask`s, **raise deprecation warning**.
+* Always eventually run with `Estimator._run(tasks: Sequence[EstimatorTask], **run_options)`.
+
+Here is a mock implementation, with missing type annotations as above:
+```python
+class Estimator(BasePrimitive):
+
+    def run(
+        self, 
+        tasks: Sequence[EstimatorTask] | EstimatorTask,
```
jyu00, line 352: A minor detail, but we'd have move `tasks` to the end for backward compatibility.

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+```python
+Estimator.run(cirucuit, parameter_values_array, observables_array)
+```
+
+This has the advantage of a simpler interface, where multiple tasks could be run 
+by invoking the estimator multiple times. The disadvantages, which we feel are significant enough to forego this simplification, are that:
+
+ 1. For real backends, the user would lose the ability to cause multiple types of circuits to be loaded into the control hardware at one time. For example, if using an estimator to perform randomized benchmarking, each circuit depth would need to be a separate job.
+ 2. It would be difficult for implementations that include mitigation to share resources between tasks. For example, if different tasks represent different trotter step counts, there would need to be a complicated mechanism to share learning resources---that are specific to the application circuits---between multiple jobs.
+
```
jyu00, line 428: Another possible migration path is to introduce a new version, e.g. `EstimatorV1`. Importing an `Estimator` will default to V1, but people can choose to import the old version.

0015-estimator-interface.md
t-imamichi: I see. Validation of inputs is not the focus of this RFC. So, I'm fine to skip it.

0015-estimator-interface.md
```diff
@@ -0,0 +1,457 @@
...
+        circuits = None,
+        observables = None, 
+        parameter_values = None, 
+        **run_options
+    ):
+        # Coerce into standard form to later allow for `tasks` to be the only non-kwarg.
+        if isinstance(tasks, QuantumCircuit) or all(isinstance(task, QuantumCircuit) for task in tasks):
+            if circuits is not None:
+                raise ValueError("Cannot mix old and new APIs")
+            circuits = tasks
```
t-imamichi, line 362: Related to Jessie's comment https://github.com/Qiskit/RFCs/pull/51/files#r1347953285.
If we put `tasks` as the first argument, we need to take care of all possibilities as follows. It might be too complicated.
```python
estimator = Estimator(...)
estimator.run([qc], [obs], [params])  # case 1
estimator.run([qc], [obs], parameter_values=[params])   # case 2
estimator.run([qc], observables=[obs], parameter_values=[params])   # case 3
estimator.run(circuits=[qc], observables=[obs], parameter_values=[params])   # case 4
```

The current code can handle case 3 and 4 correctly but fails with case 1 and 2.

As for case 1, we need to set arguments as follows.
```python
parameter_values = observables
observables = circuits
circuits = tasks
```

As for case 2, we need to set arguments as follows.
```python
observables = circuits
circuits = tasks
```

It would be a bit complicated to handle all cases properly. It would be nice to move `tasks` to the end as Jessie suggested https://github.com/Qiskit/RFCs/pull/51/files#r1347953285.

In this case, users always need to use keyword argument `tasks = ...` until we remove the legacy arguments.
There is an option to use keyword only argument for `tasks`. Then, the interface of `tasks` does not change before and after the removal of the legacy argument.
```python
# with legacy arguments
def run(self, circuits=None, observables=None, parameter_values=None, *, tasks=None, **run_options):

# after removal of legacy arguments
def run(self, *, tasks=None, **run_options):
```

If we would like to allow positional argument for `tasks`, it might be good to introduce versioning such as `EstimatorV2` as Jessie commented  https://github.com/Qiskit/RFCs/pull/51/files#r1347956614.

garrison: > It would be a bit complicated to handle all cases properly.

Another option might be to modify the signature of `run()` to use [position-only arguments](https://peps.python.org/pep-0570/).  Something like:

```python
def run(self, arg1=None, /, observables=None, parameter_values=None, *, circuits=None, tasks=None, **run_options):
    if arg1 is not None:
        if not isinstance(arg1, Iterable):
            arg1 = [arg1]
        if all(isinstance(task, EstimatorTask) for task in arg1):
            tasks = arg1
        else:
            circuits = arg1
```
with some added handling of corner cases, of course.

On one hand, this adds some complexity to the `run()` method.  However, it should work with all cases @t-imamichi mentioned above, and if the goal is for people to eventually migrate to the `Task` interface, the added complexity of handling `arg1` is a temporary stepping stone which only needs to exist for as long as the deprecation period of the old API.

0015-estimator-interface.md
ihincks: Thanks, hopefully all of these typos have been fixed in 20859ef

0015-estimator-interface.md
ihincks: Thanks, hopefully all of these typos have been fixed in 20859ef

0015-estimator-interface.md
ihincks: Indeed, I've added a clarification in 0ed2497.

0015-estimator-interface.md
ihincks: Yes. We plan to put that in a separate RFC.

blakejohnson:
> Can we make it the variance, not the standard error? i.e. it should not go to zero with increased number of shots. The user can calculate standard error given the number of shots.

I, too, would like to keep this out of scope of this RFC. But, I would argue that you want standard error of the mean as it is the thing that corresponds most closely to the error bar of the estimator. Alternatively, we could express this as a confidence interval (would have to pick the threshold... 90%?). Either definition would continue to have a well-defined meaning when using an extrapolator in ZNE.

0015-estimator-interface.md
t-imamichi: Thanks, Jim. Your solution looks nice.

ihincks:
At the moment, it seems like there is (rightly) a lot of concern about the migration path. There have been several useful suggestions about how to reorder the arguments, or whether to introduce versions. I'm going to leave all of the corresponding threads open for a bit while we work through it.

0015-estimator-interface.md
```diff
@@ -0,0 +1,502 @@
...
+One circuit, many observables:
+
+```python
+    # old API
+    estimator.run([circuit] * 4, observables=["ZIII", "IZII", "IIZI", "IIIZ"]).result()
+    >> EstimatorResult(np.array([0.1, 0.2, 0.3, 0.4]), metadata={...})
+
+    # new API (we can use a single task)
+    estimator.run((circuit, ["ZIII", "IZII", "IIZI", "IIIZ"])).result()
+    >> [ResultTask({"evs": np.array([0.1, 0.2, 0.3, 0.4]), "stds": <...>}, metadata={...})]
```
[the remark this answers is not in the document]
ihincks: Agreed, we need better names here. The dataclass thing is in the works, and will make an appearance in the next RFC that @blakejohnson mentioned below, just haven't had a chance to update this PR yet.

0015-estimator-interface.md
```diff
@@ -0,0 +1,502 @@
...
+])
+
+# wait for the result
+job.result()
+
+# get a bundle of results for every task
+>> EstimatorResult(
+>>     TaskResult<{evs: ndarray<4, 32>, stds: ndarray<4, 32>}, local_metadata>, 
+>>     TaskResult<{evs: ndarray<18>, stds: ndarray<18>}, local_metadata>, 
+>>     global_metadata
```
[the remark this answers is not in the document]
ihincks: I agree, we'll work towards having just one.

0015-estimator-interface.md
```diff
@@ -0,0 +1,502 @@
...
+
+    # new API (unique circuits always need their own tasks)
+    estimator.run([(circuit1, "ZIII"), (circuit2, "IZII"), (circuit3, "IIZI")]).result()
+    >> [
+    >>     ResultTask({"evs": np.array(0.1), "stds": <...>}, metadata={...}), 
+    >>     ResultTask({"evs": np.array(0.2), "stds": <...>}, metadata={...}), 
+    >>     ResultTask({"evs": np.array(0.3), "stds": <...>}, metadata={...})
+    >> ]
+```
+
```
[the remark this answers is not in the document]
ihincks: In one of the examples, you can see the run method being invoked as `estimator.run((circuit, ["XYZI"], [[1,2,3], [3,4,5], [8,7,6]]))`, which is already a coersion from what should more properly `estimator.run([(circuit, ["XYZI"], [[1,2,3], [3,4,5], [8,7,6]])])` aimed at reducing the number of parens that need to be typed.

However, it seems like you make be asking for even one more: estimator.run(circuit, ["XYZI"], [[1,2,3], [3,4,5], [8,7,6]])`. In my opinion, this would be a step too far, as it would fundamentally change the Python signature of the method.

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+```python
+# Broadcast single observable
+parameter_values = np.array(1.0)
+parameter_values.shape == ()
+observables = ObservablesArray(["III", "XXX", Pauli("YYY"), "ZZZ", Pauli("XYZ")])
+observables.shape == (5,)
+>> task_result.shape == (5,)
+
+# Inner/Zip
```
1ucian0, line 275:
```suggestion
# Zip
```

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+  (circuit2, observables2)
+])
+
+# wait for the result
+job.result()
+
+# get a bundle of results for every task
+>> EstimatorResult(
+>>     TaskResult<{evs: ndarray<32, 4>, stds: ndarray<32, 4>}, local_metadata>, 
+>>     TaskResult<{evs: ndarray<18>, stds: ndarray<18>}, local_metadata>,
```
woodsp-ibm, line 93: Are TaskResults here in some for of list/array corresponding to the broadcast shape?

0015-estimator-interface.md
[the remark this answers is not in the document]
woodsp-ibm: It is good to have more data fields as standard since metadata is arbitrary - in the existing Estimator it was not agreed on what stds was and was left at whatever was done, if anything, to something placed in metadata. That meant code could not easily be written to use such data since you would have to know about the specific implementation and what that provided in the metadata.

mtreinish:
We held a RFC review meeting earlier today to discuss the proposal. The recording of this meeting here:

https://ibm.webex.com/ibm/ldr.php?RCID=5835b32e58dff7d94b0522775d0e8a7a

PW: eKe5NvNQ

0015-estimator-interface.md
ihincks: Thanks. I've made the same change to the figure.

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+
+The results from each `EstimatorTask` will be array valued, and each `EstimatorTask` in the same job may have a different shape.
+Consider an `EstimatorTask` with shape `<20, 30>`, where the shape has come from the broadcasting rules discussed elsewhere. 
+This will result in a 20×30 array of real estimates.
+Moreover, we will want to return an array of standard deviations of the same shape.
+This would result in a bundle of broadcastable arrays:
+
+```python
+TaskResult({“evs”: <20, 30>, “stds”: <20, 30>}, metadata)
+```
```
jakelishman, line 346: Just writing down something I mentioned in the meeting, for posterity:

There's questions about whether standard deviation or error should be returned, but I think there's general consensus that what's returned there should have some relation to the "error bars" we'd draw on a graph, so they're representative of some sort of confidence interval.

When error mitigation is in the picture, this confidence interval is not necessarily symmetric. If nothing else, the raw estimation of the result of a single shot isn't naturally normally distributed but binomially, but in general, mitigation strategies aren't going to have their best estimates of the confidence interval be symmetric.  Giving a single number here is never going to be sufficient to describe the whole underlying multivariate distribution, so the only useful thing to return is some approximate quantity that can be given a clear, universal meaning.

The std/variance/std.error based on a univariate normal approximation is one choice there.  Another would be to define that the return value has shape `<20, 30, 2>`, with the indices being the 5-95% confidence bounds on the value (or whatever arbitrary choice of percentiles), and say that an implementer of `Estimator` is free to return more detailed statistical information in the `metadata` in a form of their choosing.  There's still no concept of the results because drawn from a multivariate distribution, so it's just choosing different places to make the approximation in.  I think a confidence interval on the marginal of each variable might be more easily defined/implemented for a general interface than "the std.err on the 'closest' normal approximation to the marginal for some implementer-defined definition of 'closest'" (or whatever).

It could also be an option to supply a couple of instance methods on `TaskResult` that take whatever uncertainty information we end up requiring of implementers, and calculate an approximation to the stddev etc.

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+
+The proposed `BindingsArray` object would support, at a minimum, the following constructor examples for a circuit with three input parameters `a`, `b`, and `c`, where we will use `<>` to denote some `array_like` of the specified shape:
+
+```python
+# a single value for each a, b, and c; 0-dimensional, and compatible with current interface
+BindingsArray([0.0, 1.0, 2.0])
+
+# a single value for each a, b, and c; 0-dimensional, and compatible with current interface
+# satisfies user wish to be able to specify values by name
+BindingsArray(kwvals={Parameter("a"): 0.0, Parameter("b"): 1.0, Parameter("c"): 2.0})
```
garrison, line 213: This line uses `kwvals`, but a below line uses `kwargs`.  Is there a difference?

At first I wanted to replace `kwvals=` with the standard python syntax for passing kwargs (`**`), but then I realized that a `Parameter` is not a valid kwarg; it has to be a string.

I do prefer a name other than kwargs, which already has a precise meaning in python.  Perhaps kwvals is better.  (It's still a non-standard abbreviation, though.)

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+BindingsArray(<50, 3>) 
+
+# specify bindings separately for each parameter, required if they have different types
+BindingsArray([<50>, <50>, <50>]) 
+
+# include parameter names with the arrays, where parameters can be grouped together in tuples, or supplied separately
+BindingsArray(kwargs={(a, c): <50, 2>, b: <50>}) 
+
+# “args” and “kwargs” can be mixed
+BindingsArray(<50, 2>, {c: <50>})
```
garrison, line 225: Not sure what the identifiers `a`, `b`, and `c` are.  Perhaps you mean the following?

```suggestion
BindingsArray(kwargs={(Parameter("a"), Parameter("c")): <50, 2>, Parameter("b"): <50>}) 

# “args” and “kwargs” can be mixed
BindingsArray(<50, 2>, {Parameter("c"): <50>}) 
```

0015-estimator-interface.md
ihincks: > This line uses kwvals, but a below line uses kwargs. Is there a difference?

`kwargs` is the typo, let me fix that while we continue to discuss.

> At first I wanted to replace kwvals= with the standard python syntax for passing kwargs (**), but then I realized that a Parameter is not a valid kwarg; it has to be a string.

There are two things that make it not a string: the ability to do things like `{(Parameter("a"), Parameter("b")): <..., 2>}`, and as you say, `Parameter` not being a string for those cases where you include an array for each parameter separately. I think the former limitation is the stronger argument for avoiding kwargs in the signature (as opposed to a dictionary argument).

I meant to bring up, so thanks for reminding me (and maybe @jakelishman would have an opinion): should these be strings or `Parameters`? This is more from a "where is qiskit going with parameters" perspective than from the issue you raised, though the issue you raised is relevant.

> Perhaps kwvals is better.

Right now it's `vals` and `kwvals`. Happy to consider alternatives.

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+    >>     ResultTask({"evs": np.array(0.3), "stds": <...>}, metadata={...})
+    >> ]
+```
+
+
+## Detailed Design <a name="detailed-design"></a>
+
+### Tasks <a name="tasks"></a>
+
+We propose the concept of a _task_, which we define as _a single circuit along with auxiliary data required to execute the circuit relative to the primitive in question_. This concept is general enough that it can be used for all primitive types, current and future, where we stress that what the “auxiliary data” is can vary between primitive types.
```
garrison, line 177: Something I mentioned in the meeting yesterday (even though the general consensus was that it is beyond the scope of this RFC):

This RFC does not mention that concept of "shots" anywhere.  If multiple circuits are submitted as multiple tasks to the same job, then they currently must have the same `shots` count (as specified in the `run_options`).  However, there are use cases (such as circuit cutting) where the results of some circuits contribute more weight to the final result than others, and so it would be desirable to run those circuits/tasks with proportional effort, some with more shots than the others.  Even after this RFC, the only way to accomplish this will be to submit the circuits in separate jobs, but this would remove the benefits of [pipelining](https://en.wikipedia.org/wiki/Protocol_pipelining) them all into the same remote procedure call.

In some uses cases, it might even be beneficial for the shots count to be adjustable on a per-parameter basis, not just a per-task basis.

In response to my comment, somebody mentioned that it might be even better for the estimator to be controlled by a desired error threshold than a shots count.  But even in that case, it would be nice to apply this at a more fine level of granularity than per-job.

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+            self,
+            circuits: Sequence[QuantumCircuit] | QuantumCircuit,
+            observables: Sequence[BaseOperator | PauliSumOp | str] | BaseOperator | PauliSumOp | str,
+            parameter_values: Sequence[Sequence[float]] | Sequence[float] | float | None = None,
+            **run_options,
+        ):
+        ...
+```
+
+We propose the migration to using tasks in two phases.
```
mtreinish, line 378: As I brought up during the meeting (but didn't get to explore fully due to time), this section covers the backwards compatibility from the estimator end user perspective. But the other axis to consider here is that we need to maintain compatibility for implementers of the base primitive classes. `BaseEstimator` provides an abstract base class that implementers need to subclass to provide an estimator implementation. To provide backwards compatibility it's a requirement to have the interface changes work in a manner that's compatible with **existing** implementations of `BaseEstimator`. Not all implementations of the estimator are part of qiskit itself, and we need to provide a migration path for all downstream implementations to enable using the new interface. Also any internal changes do not interfere with the operation of existing primitives in otherwise previously valid input. 

We can have previously implementations error on task based input (as that wasn't supported before), but that is essentially versioning the interface implicitly and it becomes an exercise for users to negotiate which estimators support which features. In such cases I feel it's much better to be explicit about the features supported via versioning, So, I think there are really 2 paths forward here, either explicitly versioning the `BaseEstimator` interface (i.e. adding a `BaseEstimatorV2` based on what's described in this RFC) and documenting the differences and the end user migration path for consumers of estimators. Then the user can determine which version of the estimator there using and and adjust their usage. Or alternatively we add a parallel internal dispatch mechanism to `BaseEstimator` for the new task based interface so that if a user submits a task style input that is converted into the format existing implementations will support:

https://github.com/Qiskit/qiskit/blob/fe3bffbc85a4fb57dc62ba04fa7f11bcefb5d75f/qiskit/primitives/base/base_estimator.py#L196-L202

Basically this means adding a new optional method like `def _new_run()` which accepts task based input, in the case `_new_run()` is not defined in an implementation `BaseEstimator` will convert the new style input into the old style (likely via a lot of copying) and pass that to `_run()`. This way if end users submit a task based input to `.run()` it will work correctly (albeit inefficiently) even for existing implementations of `BaseEstimator`.

woodsp-ibm: For `_run()`, for reference:
 * Qiskit/qiskit#11057

0015-estimator-interface.md
ihincks: I expect that this would be possible as an additive change, either by allowing another field to `EstimatorTask` that accepts precision, or in the `ObservablesArray`. Precision could possibly be array-based if you needed different precision for different estimates.

0015-estimator-interface.md
ihincks: Okay, I think pure strings is probably a bit cleaner too, I'll make the changes in that direction.

I don't want to lose the ability `{("a", "b"): <..., 2>}`, and my personal preference is to not have too many ways to do the same thing, so I'm in favor of not introducing direct unpacked at this time, with the possibility of introducing it as an enhancement in the future.

0015-estimator-interface.md
ElePT: I was just discussing this with @Cryoris, that we could maybe avoid the overall interface versioning by having a `def _run_v2()` (or `def _new_run()`), but https://github.com/Qiskit/qiskit/pull/11057 would block that potential path.

0015-estimator-interface.md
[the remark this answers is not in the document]
ElePT: It would change the signature, the same way it did with the original interface's signature, and maybe this use-case is not a priority anymore. But I wanted to make sure you were aware that this actually the most common use of primitives in most high level applications (which are based on variational algorithms, such as VQE) and that this was feedback that we received and ended up being acted on in the past.

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+For example, 1D list of 10 objects (typically numbers) is a vector and has shape `(10,)`; a matrix with 5 rows and 2 columns is 2D and has shape `(5, 2)`; a single number is 0D and has an empty shape tuple `()`; an `nd-array` with shape (20, 10, 5) can be interpreted as a length-20 list of 10×5 matrices.
+
+We propose the constraint that a `BindingsArray` instance and an `ObservablesArray` instance that live in the same task must have shapes that are broadcastable, in which case, the `Estimator` promises to return one expectation value estimate for each element of the broadcasted shape.
+
+Here are some examples of common patterns expressed in terms of array broadcasting, and their accompanying visual representation in the figure that follows:
+
+```python
+# Broadcast single observable
+parameter_values = np.array(1.0)
+parameter_values.shape == ()
```
levbishop, line 270: Can I suggest not allowing the behavior of implicitly appending extra (1) dimensions for broadcasting? I find this aspect of matlab/numpy array-handling to be error-prone and I'd prefer explicit dimensions in all cases, whether prepending (where it is already required) or appending (where it is currently optional):
yes (values.shape, observables.shape -> results.shape):

> (5,) (1,) ->(5,)
> (5,7) (5,1) -> (5,7)
> (5,7) (1,7) -> (5,7)

No:

> (5,) () -> (5,)
> (5,7) (5,) -> (5,7)

Providing input in this form is explicit using np.newaxis, which just fiddles with array metadata doesn't require memcpy or anything:

```
In [9]: x.shape
Out[9]: (2, 3)

In [10]: x[:,np.newaxis,:,np.newaxis].shape
Out[10]: (2, 1, 3, 1)
```

blakejohnson (approved):
Approving from the lens of the desired interface, including:
* tasks (we can discuss name of this elsewhere)
* broadcasted shapes
* `ObservablesArray`
* `BindingArray`
* bumping up error bars into the required part of the `TaskResult`
 
The issue of migration path is important, but not in scope of my review.

0015-estimator-interface.md
ihincks: That's interesting. By the way, the numpy convention is to right align (I blundered in some version of this document), so `(5,7) (5,) -> (5,7)` should actually be `(5,7) (7,) -> (5,7)`. But this serves your point.

Despite this, I do think there's value in being able to say "it's just like numpy", and I also think there's value in being able to provide 0-D values to broadcast. Maybe others want to weigh in.

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+
+1. _Performance._ Given that `QuantumCircuit` hashing should be avoided and circuit equality checks can be expensive, we should move away from an interface that necessitates performant primitive implementations to go down this path at all. As one example, qubit-wise commuting observables like `"IZZ"` and `"XZI"` can both be estimated using the same simulation/execution, but only if the estimator understands that they share a base circuit. As a second example, when the circuits need to be serialized before they are simulated/executed (e.g. runtime or pickling-for-multi-processing), it puts the onus on the primitive implementation to detect circuit duplications.
```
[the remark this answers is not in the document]
ihincks: Your comments about the inadequacies of what the primitives do between calls to `run()` are fair, but out of scope; I'd like to see them addressed in a future work. This can be done without compromising the signature proposed here.

This RFC is strictly about the API of the `run()` method, and so really can't make promises about an implementation's internal behaviour. It does, however, address the caching question within a single call. It seems like this won't have much value to your particular workloads, but it does to others, and in particular it has tremendous value to how we get to implement things in deeper areas of our stack.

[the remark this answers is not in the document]
woodsp-ibm: My comment was less about the run signature more about this below, which was talking about expectation of the change around implementation

> hashing should be avoided and circuit equality checks can be expensive, we should move away from an interface that necessitates performant primitive implementations to go down this path at all

If the primitive is just used for one run then the onus would be on the user not to supply duplicates in the inputs anywhere for this to apply (i.e. no duplicate checks needed). If they do have duplicates in input then its down to implementation as to what they do. Same across runs - but without any equality/check caching, performance across runs is going to take a hit and if one has it for that then you have it anyway. It was less clear to me that changing the interface would have any implementation change in that regard as a result of this RFC. The change would make it theoretically possible but for all practical purposes, unless one wanted to take a performance hit across runs - the Session and the primitives was all about improving iterative performance - it seems less likely in that regard. 

But yes,as you say, the RFC is about the interface not the implementation so it maybe a step towards doing the following, at some point in the future if further RFC(s) cover additional changes to allow this to really make this happen.

> we should move away from an interface that necessitates performant primitive implementations to go down this path (_hashing and circuit equality checks) at all

0015-estimator-interface.md
ihincks: There is one TaskResult for each Task. Each TaskResult is shaped like the corresponding task.

woodsp-ibm: I think my question was more that the TaskResults seemed to be individual parameters to the EstimatorResult object, followed by some global metadata. Below estimator.run (line 164) shows run returning a list of ResultTasks. The input to the run() just above is a list of tuples- the two "tasks". Maybe this is more illustrative here than actual in that it shows two things as well, and maybe they end up in a list or some Iterable part of result.

blakejohnson:
Please, let's close this in the next 24 hours.

0015-estimator-interface.md
ihincks: I propose to hash it out in the PRs.

levbishop: I don't have a strong opinion either way and agree we can hash it out in PRs just wanted to raise the question

levbishop (approved):

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+>> EstimatorResult(
+>>     TaskResult<{evs: ndarray<32, 4>, stds: ndarray<32, 4>}, local_metadata>, 
+>>     TaskResult<{evs: ndarray<18>, stds: ndarray<18>}, local_metadata>, 
+>>     global_metadata
+>> )
+```
+
+The above example shows various features of the design, including combining large numbers of observables with large numbers of parameter value sets for a single circuit.
+
+The detailed targetted signature, following the [deprecation period](#migration-path),
```
jlapeyre, line 100:
```suggestion
The detailed targeted signature, following the [deprecation period](#migration-path),
```

0015-estimator-interface.md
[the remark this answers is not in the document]
ihincks: Thanks for emphasizing this point. We'll do our best to have strong examples in this regard.

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+One circuit, many observables:
+
+```python
+    # old API
+    estimator.run([circuit] * 4, observables=["ZIII", "IZII", "IIZI", "IIIZ"]).result()
+    >> EstimatorResult(np.array([0.1, 0.2, 0.3, 0.4]), metadata={...})
+
+    # new API (we can use a single task)
+    estimator.run((circuit, ["ZIII", "IZII", "IIZI", "IIIZ"])).result()
+    >> [ResultTask({"evs": np.array([0.1, 0.2, 0.3, 0.4]), "stds": <...>}, metadata={...})]
```
itoko, line 141:
```suggestion
    >> [TaskResult({"evs": np.array([0.1, 0.2, 0.3, 0.4]), "stds": <...>}, metadata={...})]
```

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+One circuit, one observable, many parameter value sets:
+
+```python
+    # old API
+    estimator.run([circuit] * 3, observables=["XYZI"] * 3, parameter_values=[[1,2,3], [3,4,5], [8,7,6]]).result()
+    >> EstimatorResult(np.array([0.1, 0.2, 0.3]), metadata={...})
+
+    # new API (we can use a single task)
+    estimator.run((circuit, ["XYZI"], [[1,2,3], [3,4,5], [8,7,6]])).result()
+    >> [ResultTask({"evs": np.array([0.1, 0.2, 0.3]), "stds": <...>}, metadata={...})]
```
itoko, line 153:
```suggestion
    >> [TaskResult({"evs": np.array([0.1, 0.2, 0.3]), "stds": <...>}, metadata={...})]
```

0015-estimator-interface.md
```diff
@@ -0,0 +1,500 @@
...
+    # old API
+    estimator.run([circuit1, circuit2, circuit3], observables=["ZIII", "IZII", "IIZI"]).result()
+    >> EstimatorResult(np.array([0.1, 0.2, 0.3]), metadata={...})
+
+    # new API (unique circuits always need their own tasks)
+    estimator.run([(circuit1, "ZIII"), (circuit2, "IZII"), (circuit3, "IIZI")]).result()
+    >> [
+    >>     ResultTask({"evs": np.array(0.1), "stds": <...>}, metadata={...}), 
+    >>     ResultTask({"evs": np.array(0.2), "stds": <...>}, metadata={...}), 
+    >>     ResultTask({"evs": np.array(0.3), "stds": <...>}, metadata={...})
```
itoko, line 168:
```suggestion
    >>     TaskResult({"evs": np.array(0.1), "stds": <...>}, metadata={...}), 
    >>     TaskResult({"evs": np.array(0.2), "stds": <...>}, metadata={...}), 
    >>     TaskResult({"evs": np.array(0.3), "stds": <...>}, metadata={...})
```

itoko (commented):
fix minor typos (ResultTask -> TaskResult)

ihincks:
Thanks everyone for the thoughtful discussion about whether to version the interfaces, or whether to migrate the current signature. We have, in the end, decided that the pros of versioning the interface win.

0015-estimator-interface.md
```diff
@@ -0,0 +1,414 @@
...
+  2. Suppose we want to return some indication of which estimates came from the same physical circuit.
+
+## Migration Path <a name="migration-path"></a>
+
+We need to remain backwards compatible with the existing interface to adhere to the [Qiskit Deprecation Policy](https://qiskit.org/documentation/deprecation_policy.html).
+
+In summary, we propose the following strategy: explicitly version the `Estimator` interface (i.e. add a `BaseEstimatorV2` as described in this proposal) and document both the differences and the end user migration path for consumers of estimators. The user can determine which version of the estimator to use, and existing implementations can be left intact.
+
+Similar to `Backend`, we will have a `BaseEstimator`, `BaseEstimatorV1` equal to the existing `BaseEstimator`, and `BaseEstimatorV2`, as described in this proposal.
+We will maintain backward compatibility by defaulting the import of `Estimator` to `EstimatorV1` (i.e. existing) for the duration of the deprecation period.
```
mtreinish, line 370: Which `Estimator` are you talking about here? I assume `qiskit.primitives.Estimator`, but I know of at least three classes named `Estimator` that implement `BaseEstimator` in different repositories.

Regardless of which one is being referred to, one other option here that might make it easier for users, is basically what `qiskit-ibm-provider` does for its `BackendV2` implementation. The backends from `qiskit-ibm-provider` implement the `BackendV2` interface but also provide a compatible interface with most of `BackendV1` (the notable exception being `backend.name` which is an attribute on V2 and a method on V1). So users have the freedom to use both interfaces for an existing implementation. It also makes future deprecation a little easier because it's easier to warn on legacy access patterns and inputs. But it's also totally manageable with parallel implementations, just using a future warning when `Estimator` is instantiated and you can use a `__future__` import to change `Estimator` over to `EstimatorV2`.

1ucian0 (approved):

mtreinish (approved):
With a decision being made to use explicit versioning on the public interface as a migration strategy I'm happy with the proposal now. Thanks for all the hard work sorting through everyone's comments to come up with a concrete and workable plan.

I left 2 small comments inline about some of the migration path. Neither are critical for this RFC, the converter class I think is a good idea and some we can add independently after `BaseEstimatorV2` has been defined. The other is more of a question for clarification and idle musings on migration strategies for implementers of Estimators (which is largely out of scope for this document, but useful to discuss while formulating this plan).

ElePT (approved):

1ucian0:
I think the only two open discussion are:
 * https://github.com/Qiskit/RFCs/issues/54
 * https://github.com/Qiskit/RFCs/issues/55
Let’s move those conversation to their issues.

PR #56, opened 2023-10-31
Updated Sampler

ihincks:


0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+|                   | Blake Johnson (blake.johnson@ibm.com)        |
+|                   | Chris Wood (cjwood@us.ibm.com)               |
+| **Submitted**     | 2023-10-03                                   |
+| **Updated**       | YYYY-MM-DD                                   |
+
+
+## Summary <a name="summary"></a>
+
+This RFC proposes an interface for the next version of the `Sampler`.
+The changes proposed here mirror those accepted in [(the paired Estimator inteface RFC)[https://github.com/Qiskit/RFCs/pull/51]](https://github.com/Qiskit/RFCs/pull/51), where the arguments of the `Estimator.run()` signature were transposed to accept, instead of three iterable arguments, an iterable of "tasks", with each task comprising a circuit and auxiliary execution information.
```
garrison, line 17:
```suggestion
The changes proposed here mirror those accepted in [the paired Estimator interface RFC](https://github.com/Qiskit/RFCs/pull/51), where the arguments of the `Estimator.run()` signature were transposed to accept, instead of three iterable arguments, an iterable of "tasks", with each task comprising a circuit and auxiliary execution information.
```

blakejohnson (approved):

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+        """
+```
+
+### Measurement Levels
+
+`backend.run` has the notion of measurement levels which specify whether measurement instructions cause traces (not generallly implemented), complex phases, or discriminated bits to be returned.
+Eventually, measurement levels should be specified inside of a circuit itself by measuring into complex valued registers, as described in the OpenQASM 3 specification.
+This would effectively enable a mix-and-match between different measurement levels, and also allow a circuit to determine its own output types.
+
+However, before this becomes available, in order to acheive feature parity with `backend.run`, we propose implementing this as a flag (similar to `backend.run`) whose value changes the type of all bit-array outputs from `BitArray`s to NumPy arrays with `dtype=np.complex128`.
```
ajavadia, line 250: Originally the plan was to have a different primitive, called "Discriminator", which would return IQ values for measurements (and then enable different options for how we do kernels etc. as a form of error suppression at this level). 
I think it's fine to fold this into the Sampler, especially since it becomes compatible with QASM3 and allows mix-and-match of different measurements, but maybe this should be called out that building a Discriminator is not planned anymore.

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+>>     SamplerTaskResult(
+>>         DataBundle(
+>>             alpha=BitArray(<num_samples=1024, num_bits=127, shape=()>),
+>>             beta=BitArray(<num_samples=1024, num_bits=10, shape=()>),
+>>             gamma=BitArray(<num_samples=1024, num_bits=15, shape=()>)
+>>         ), 
+>>         <metadata>
+>>     ), 
+>>     metadata=<global_metadata>
+>> )
```
ajavadia, line 88: Two questions about the returned result datastructures:
- Is this also compatible if, in the future, we want to enable streaming of results? do we need to change the datastructures at all?
- similar to Estimator, I think it would be good to add a field here which is the confidence interval, and it gets smaller with more shots.

ajavadia (commented):
overall direction and philosophy looks good to me

0017-sampler-interface.md
ihincks: 1) I imagine that a streaming interface would look substantially different, and would require new tooling. The emphasis here is on data contiguity for fast local operations. I think it would probably be best to treat that with separate tools.

2) I'm wondering what the confidence interval would be of? The observations are the observations

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+# flip all of the bits in the gamma according to some other compatibly-shaped bit-array
+flips = BitArray(my_flips, num_bits=15)
+flipped_gamma = result.data.gamma ^ flips
+```
+
+If a circuit had a non bit-array outcome, it would use a different type in place of `BitArray`.
+We suggest just using `numpy.ndarrays` with the closest appropriate `dtype` for non-bit-array outputs. 
+The shape of such an array would be `(*task_shape, num_samples, *output_register_shape)`.
+For particular types, like `angle`, if it is for some reason decided that no NumPy type is satisfactory,
+we can add more containers analagous to `BitArray`.
```
jakelishman, line 118: This is an area where we expect both Qiskit and `Sampler` implementers to be evolving in the near term.  In particular, specific `Sampler` implementations are likely to start supporting different OQ3 output types (bit array vs others) at different rates, and Qiskit will expand its capability to represent outputs, which will then take time for sampler implementations to adapt to.

Is it worth considering how we can make the `SamplerV2` (and `EstimatorV2`, I suppose) easier to expand in the future?  One possible example, having an instance expose a "supported features" attribute, similar to the IBM backends of today, which continues values that have precisely defined semantics (defined _only_ by the base interface) for what a user can expect this sampler to support?  As a very very rough idea, a "feature" could be `"IMPLEMENTATION_LEVEL_1"` and this could mean "circuits that contain a single input qubit register and a single output classical register of the same width, and contain only a sequence of gate calls without classical conditions followed by a sequence of measurements".  This kind of built-in versioning could give us a bit more scope to expand the capabilities of the `Sampler` in the future, and to define the required output types for more dynamic circuits later.

(This isn't something that seems obviously correct in my head at all, I'm just thinking about trying to avoid being in another breaking-change `SamplerV3` situation next year.)

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+            ValueError: If no ``idx`` is given and this task result is more than 0-D.
+        """
+```
+
+### Measurement Levels
+
+`backend.run` has the notion of measurement levels which specify whether measurement instructions cause traces (not generallly implemented), complex phases, or discriminated bits to be returned.
+Eventually, measurement levels should be specified inside of a circuit itself by measuring into complex valued registers, as described in the OpenQASM 3 specification.
+This would effectively enable a mix-and-match between different measurement levels, and also allow a circuit to determine its own output types.
+
```
jakelishman, line 249: IBM's backends have the concept of measurement levels in `backend.run`, but the interface itself doesn't, and Aer for example doesn't support it (or at least certainly not with the name `meas_level`).  It's handled kind of specially in Qiskit's `Result` class, but the input fields it refers to are inherited from the old deprecated `Qobj`, not `BackendV1` or `BackendV2`.

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+However, before this becomes available, in order to acheive feature parity with `backend.run`, we propose implementing this as a flag (similar to `backend.run`) whose value changes the type of all bit-array outputs from `BitArray`s to NumPy arrays with `dtype=np.complex128`.
+
+```python
+class MeasLevel(enum.Enum):
+    # TRACES = 0 unsupported
+    PHASES = 1
+    BITS = 2
+```
```
jakelishman, line 257: Since this is an IBM extension, does it need to be in the `Sampler` interface definition?  There's no need for the flag for feature parity given the Aer implementation, for example.

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+ * _Has no sample weights._ The current sampler attaches real weights summing to one to all samples. In effect, it returns an estimate of a (quasi)-distribution instead of samples from the distribution. This proposal eliminates this feature and returns only samples, and without weights. The rationale is to make the `Sampler` into more of a streamlined circuit executor, paring down all frills.
+
+
+## User Benefit <a name="user-benefit"></a>
+
+Users will be able to transition from `backend.run` to `sampler.run` without any compromises in execution efficiency. They will be able to run dynamic circuits without having to specify special flags, or invoke a special program. They will get their outputs in a format where they don't need to manually separate bitstrings from different registers.
```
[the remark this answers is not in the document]
blakejohnson: See my proposal in https://github.com/Qiskit/qiskit-ibm-provider/issues/640 for removing most of these. Whatever remains can be bundled into some execution options.

jakelishman (commented):
Thanks for this, Ian.  I think there's a few things in here that treat IBM-specific extensions as if they are part of the `backend.run` general interface, which might want removing / reducing.

On a larger note, I'm wondering how / if we can ensure that `Sampler` (and to a degree, `Estimator`) are expandable interfaces, potentially with some form of version negotiation, so we don't end up needing to make a `SamplerV3` in the future that might return different types.  This maybe is a concern too far and too late at this stage, though.

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+alpha = ClassicalRegister(127, "alpha")
+beta = ClassicalRegister(10, "beta")
+gamma = ClassicalRegister(15, "gamma")
+circuit2 = QuantumCircuit(QuantumRegister(127), alpha, beta, gamma)
+...
+<insert gates, measure into the registers, apply dynamic circuit features, etc.>
+...
+
+# invoke a sampler
+sampler = Sampler(..., shots=1024)
```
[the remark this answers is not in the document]
blakejohnson: I don't think the intention is for `shots` to be a required argument in the `Sampler` constructor.

0017-sampler-interface.md
blakejohnson: You could store such information in the task result metadata.

0017-sampler-interface.md
ihincks: It seemed pertinent given the claim that we'd like to phase out `backend.run`, and also because we're pointing out an explicit exception to the policy "we obey program outputs".

blakejohnson: In some ways, it is an OpenPulse extension rather than just an IBM extension. This whole section of the proposal comes from trying to build a bridge toward a future state where Qiskit could represent measurements that return `complex[float[64]]` or `waveform[...]` types. When we have proper support for those types and associated measurement instructions, then this flag could disappear entirely.

Point taken that this _could_ be relegated to IBM's implementation of the Sampler.

jakelishman: It's not clear to me that the IBM-only extension of `meas_level` should get such special treatment in the base interface (especially since it wasn't part of the base interface before), and it's really quite worrying that it fundamentally changes the output type of the run method.  Having it as a required option means that all Samplers now need to support it, including those based on backends that didn't previously support it.

Is there a way to go abstractly from a raw-shots return value to the higher measurement-level return value?  If so, then can `meas_level` be a client-side postprocessing step instead of needing quite serious special-casing (including making return types inconsistent and breaking one of the core policies of Sampler) throughout the interface?  If not, then this feels like something that should be outside the remit of a Sampler.  An interface is something we need to live with for a long time if we want to drive adoption of it both from users and from other hardware vendors.  Including this, if it can't be abstracted, would pass on a lot of pain to hardware vendors, including meaning that we'd need to make even in-house projects like Aer and Terra's internal primitives newly support it.

jakelishman: > When we have proper support for those types and associated measurement instructions, then this flag could disappear entirely.

If it's in the base interface, it can't disappear easily from the interface; if `SamplerV2` defines that `run` takes it, then we can't remove it from future `SamplerV2`s without breaking user code that uses higher measurement levels, and syncing with other implementers of Sampler.

---

(My "It's not clear to me that the IBM-only ..." comment was written before I saw your comment, Blake - the core of my argument is the same though if it's an OpenPulse or IBM extension, because in practice it's only `IBMProvider` backends that have the field right now.)

ihincks: I think there's been a misunderstanding, likely arising from how it's written. This section exists to allay fears that IBM-only implementations won't have access to meas_level=1 at all. It is not making a comment on the base implementation, and the argument `meas_level=...` was in reference to how we currently allow run() to overwrite options temporarily. I appreciate this latter point is confusing because the three RFCs have said slightly different things on this front, but that is just a side-effect of writing four simultaneously.

The idea that an option value is capable of changing the output type should be treated as a separate issue. The rationale is that we are explicitly accepting a concession that users who want to use that flag are giving up portability during this intermediate period where we can't have the things we want.

jakelishman: Given that it's explicitly giving up portability, it might be cleaner to remove all mention of it from the RFC (which by my understanding is of the abstract interface, not specifically for "IBM's Sampler as a replacement of IBM's `backend.run` - disagreement on this might be where the misunderstanding is coming from), and potentially even have the IBM Sampler have a completely separate `run_with_meas_level` function, so that we don't have to give up the strong typing of the base interface.

We'll already be giving up portability by saying that "`meas_level`" can affect the output types, so it feels safer to me to have that be a (temporary) IBM extension that uses a completely different method, which really drives home to users that this is not portable and is a stop-gap helper transition.

0017-sampler-interface.md
blakejohnson: I think the proposal is you continue using continuous memory to store other types, but that in the future we may want more convenience wrappers to present that data in a usable/ergonomic fashion.

You comment about allowing implementations to declare feature support is interesting, but I think it requires its own independent design process. So, either it becomes something we can add-on to SamplerV2 as an enhancement, or it will have to wait to SamplerV3.

jakelishman: My aim is mostly to avoid needing a `SamplerV3` at all, but pragmatically, it may be too late in the design process to avoid that now; we'd risk doing it wrong if we rushed it in too fast, so it might be better to bite the bullet and know we'll have to do it again in a year's time.

The general principle is around defining expandable interfaces with version negotiation.  Terra's `UnitarySynthesis` plugins do this, for example - we say that all attributes whose names start `supports_` are reserved names, and if they're not present, they default to `False`.  Then when we add new capabilities that might want to be abstracted over, we define a new "feature flag" with particular semantics, and plugins can add it slowly over time, without us needing a new interface.

It's not a silver bullet by any means, and the case of a unitary-synthesis compiler plugin is easier than `Sampler`, but it's the general principle I was getting at.

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+ * _Is a "`memory=True`" exection model._ We will break from the tradition of using a mapping structure as the default container type for circuit exection. Instead, results will always be returned in a format where one output axis iterates over shots, and maintains execution order. The motivation is that, in a program whose total output size is more than 100 qubits, due to gate and measurement noise, the chances that any two output values are equal to each other is too small to make special accomodation for via a hash map. When the output size is smaller than this loose threshhold, so that output value collisions are likely, the data will take up less space anyways. Container objects will contain methods returning count-like structures to ease migration.
+ * _Has no sample weights._ The current sampler attaches real weights summing to one to all samples. In effect, it returns an estimate of a (quasi)-distribution instead of samples from the distribution. This proposal eliminates this feature and returns only samples, and without weights. The rationale is to make the `Sampler` into more of a streamlined circuit executor, paring down all frills.
```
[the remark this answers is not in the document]
blakejohnson: If an implementation that does those things is willing to pretend that it returns just samples from the output, then I guess so. However, I think the spirit of the proposal is to recognize those techniques as non-scalable and therefore out of scope of the Sampler. Someone willing to accept finite-size limitations from ROM can construct a quasi-prob distribution client-side.

0017-sampler-interface.md
blakejohnson: Sure, though we never published a Qiskit roadmap toward other primitives.

0017-sampler-interface.md
blakejohnson: That sounds like a useful pattern. But like you say, I think we have run out of time to add that to the Primitive interfaces in this iteration.

jlapeyre:
My understanding is that primitives are meant to be universal interfaces. But references to other "tooling" above https://github.com/Qiskit/RFCs/pull/56#discussion_r1377875689 make this less clear to me.  The particular concern I have is that sampling a large number of shots from a modest number of qubits using a simulator might be inefficient to impossible without returning results in an associative array. Is there a way to accommodate this situation within this proposal. Or would that require completely separate tooling?

blakejohnson:
@jlapeyre that comment is in reference to streaming interfaces, which are out of scope of this RFC.

> The particular concern I have is that sampling a large number of shots from a modest number of qubits using a simulator might be inefficient to impossible without returning results in an associative array. Is there a way to accommodate this situation within this proposal?

It is no more difficult than handling the same number of shots from hardware.

jlapeyre:
> It is no more difficult than handling the same number of shots from hardware.

Yes, in principle. I used a simulator as an example because it's easy to quickly generate with a simulator more shots than you can store. Anyone currently doing experiments like this won't be able to use this interface.

Now that I think about it, if I understand you correctly, I agree that rather than special casing accumulating shots in a hash map, it's better handled as part of a more general streaming or "online" tool that can handle custom filters. And I see it's been suggested that this is out of scope: https://github.com/Qiskit/RFCs/pull/56#discussion_r1377875689

I'd consider making the streaming interface the fundamental one and provide a `collect` function to write to an array. Then build something like this proposal on top of that. But it looks like this is too late for this version.

jyu00 (approved):

0017-sampler-interface.md
blakejohnson: I will note that the IBM hardware discriminators don't return enough information to provide this information in general. You need to activate the `meas_level=1` interface in `backend.run` to have access to that data today.

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+ * _Naturally allows multiple parameter value sets._ A common use case for many users is running the same parametric program with many parameter value sets in the same `run()` invocation. These changes will make it easier for implementations to do this efficiently with late-stage binding. Efficient invocation of the same parametric circuit between multiple `run()` invocations is out of scope of this RFC.
+ * _Is a "`memory=True`" exection model._ We will break from the tradition of using a mapping structure as the default container type for circuit exection. Instead, results will always be returned in a format where one output axis iterates over shots, and maintains execution order. The motivation is that, in a program whose total output size is more than 100 qubits, due to gate and measurement noise, the chances that any two output values are equal to each other is too small to make special accomodation for via a hash map. When the output size is smaller than this loose threshhold, so that output value collisions are likely, the data will take up less space anyways. Container objects will contain methods returning count-like structures to ease migration.
```
jlapeyre, line 27:
```suggestion
 * _Is a "`memory=True`" execution model._ We will break from the tradition of using a mapping structure as the default container type for circuit execution. Instead, results will always be returned in a format where one output axis iterates over shots, and maintains execution order. The motivation is that, in a program whose total output size is more than 100 qubits, due to gate and measurement noise, the chances that any two output values are equal to each other is too small to make special accomodation for via a hash map. When the output size is smaller than this loose threshhold, so that output value collisions are likely, the data will take up less space anyways. Container objects will contain methods returning count-like structures to ease migration.
```

0017-sampler-interface.md
```diff
@@ -0,0 +1,282 @@
...
+            ValueError: If no ``loc`` is given and this task result is more than 0-D.
+        """
+```
+
+### BitArray
+
+To store lots of bit-array outcomes, to enable fast iteration over bitstrings, and to enable bitwise operations, we propose a container that has a contiguous memory model over the entire shot-loop, where 8-bits are packed to the byte.
+For this purpose, we choose to use a NumPy `uint8` array because, unlike `bytes` or `bytearray`, it has convenient methods for slicing and indexing multiple dimensions.
+Its buffer can be directly accessed for those not wishing to use NumPy, and, as seen both in the previous section and also below, we can include convenience methods for converting to, for example, a `Counts` object.
+We choose a convention where the second last axis is over samples (shots), and the last dimension has a big-endian byte order which describes a single value of the output bit-array.
```
jlapeyre, line 184: Was this the intended meaning? ...
```suggestion
We choose a convention where the second to last axis is over samples (shots), and the last dimension has a big-endian byte order which describes a single value of the output bit-array.
```

0017-sampler-interface.md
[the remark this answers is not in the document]
ajavadia: I think M3 readout error mitigation was ostensibly scalable. And I remember one reason we said users should change their programs to use Primitives rather than `backend.run()` was that they get access to error mitigation done server side. In the case of readout error mitigation, it is much easier for the runtime to hold and update the readout error model, than for each user to characterize this separately.

I think the data type returned from "mitigated counts" might not be compatible with pure samples as it's not a string of `0` and `1`. It might be possible for those methods to reconstruct a (quasi)-probability distribution, and then resample from that to generate "fake" samples and pretend those are the real samples. I don't know if this is the correct way of going about this. 

I don't want to hold up the RFC more, but I think this requires some more thought.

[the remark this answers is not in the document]
blakejohnson: @ajavadia there very well may be users that want an interface on quasi-probability distributions. However, the methods we have today don't scale to utility-scale examples. I would happy to discuss ways to resolve that in a separate issue.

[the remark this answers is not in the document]
ihincks: Imagine a device on 1000 qubits where each qubit has a 0.01 readout error. Consider a circuit that does nothing but measure each qubit once, into a classical register on 1000 bits. In 100,000 shots, the chances of observing the all-0 bitstring are about 4e-5, and the chances of observing the all-0 bitstring in 10K shots at least once are about 35%. Consider that these numbers go quickly to 0 as a function of both circuit width and depth, keeping instruction error fixed.

M3 doesn't change the observed bitstrings, it only re-weights their likelihood away from 1/N. So in effect, the above is showing that, asymptotically, M3 is guaranteed to report a quasi-prob distribution that is disjoint from the circuit's ideal distribution. One might be able to argue that this is okay, because one is ultimately not interested in length-1000 bitstrings, but small marginals thereof. This would require careful messaging.

Keep in mind also that quadratic scaling in circuit size is a problem in practice, because eventually post-processing

> it is much easier for the runtime to hold and update the readout error model, than for each user to characterize this separately

Are there other existing readout mitigation methods, besides M3, where this actually works? Because the other ones I know about explode to 2^n in memory. You can get around this by only doing mitigation on marginalized subsystems, but this kind of thing makes more sense client side, because otherwise you need to introduce semantics about how and where to marginalize, and lose the ability to change your mind having taken the data.

> returned from "mitigated counts" might not be compatible with pure samples as it's not a string of 0 and 1

We considered adding a field to the result bundle called `weights` which would give real weights to the corresponding samples. This would effectively make the output equivalent to a multimap, where the keys are some index of the sample dimension, and the values are the weights. This could be added in the future if a mitigation procedure for which it makes sense becomes available.

PR #74, opened 2024-06-21
Sparse observables with extended alphabets

jakelishman:
### Summary

This describes a new `SparseObservable` object that would be added to `qiskit.quantum_info` for use in the `Estimator` interface, to address two major problems:

* the Pauli alphabet of `SparsePauliOp` makes it impossible to efficiently represent all operators that can be efficiently measured by hardware.
* as device qubit counts scale up, the number of non-identity terms in observable terms does not necessarily do so, making the `SparsePauliOp` representation inefficient.

### Scope

This design doc is just about what the new class will look like.  A key component of that is making sure that it _can_ be used in `Estimator` implementers, but the exact mechanisms of how the public `Estimator` interface evolves to accept this are out-of-scope of this document.

### DIFF
--- a/####-sparse-observable.md
+++ b/####-sparse-observable.md
@@ -223,9 +223,11 @@ class SparseObservableView:
     boundaries: Array[np.uintp]
 ```
 
-where `Array` is any Python object capable of being a memory view onto 1D contiguous C-stored data (not Python objects).
-This can be contiguous 1D Numpy arrays or simple `memoryview` wrappers round the Rust pointers, for example.
-Most likely, using Numpy arrays will be most convenient for users.
+where `Array` is a Python object capable of being a memory view onto 1D contiguous C-stored data (not Python objects).
+To allow `SparseObservable` to be growable in-place (which in turn allows the `+=` operator to be efficient), this will be a custom Rust-space view that implements the Numpy `__array__` protocol and allows direct reading and writing.[^1]
+
+[^1]: It has to be a custom type that can handle the case of a user holding onto a reference to the view object and then performing an operation that re-allocates the Rust-space buffers (like a grow-in-place that exceeds the available capacity).
+
 This view form will be available to enable serialisation from Python space.
 
 A enumeration `SingleBitTerm` will be available in Python space that provides named access to the integer values for each item of `s_terms`.
--- a/####-sparse-observable.md
+++ b/####-sparse-observable.md
@@ -80,7 +80,7 @@ There are several use-cases for observables throughout the entire life cycle of 
 2. All observables for an estimator pub / batch of pubs must exist simultaneously in memory on the client side and be transmitted over the wire, so must not use excessive memory.
 
 3. An estimator primitive (or circuit-knitting toolbox, etc) must be able to efficiently extract all the terms that correspond to unique measurements that must be made, across the entire broadcast `EstimatorPub`.
-   This is not necessarily the same set of mathematical terms as appear in a single operator, and the available groupings are dictated not just be individual mathematical operators or the `ObservablesArray`, but also in how it broadcasts against the `BindingsArray`.
+   This is not necessarily the same set of mathematical terms as appear in a single operator, and the available groupings are dictated not just by individual mathematical operators or the `ObservablesArray`, but also in how it broadcasts against the `BindingsArray`.
 
 4. After the measurements are made in hardware execution, the values of the observable expectations must be recalculated after.
    This means that the abstract observable should be able to easily and efficiently map back to the measurement terms needed for each of the individual abstract terms.
@@ -97,8 +97,7 @@ The operations that must be efficient for use-case 3 are also quite different to
 > This will be for users' benefit, and will be the "natural" input format of `Estimator`.
 
 A new class, `qiskit.quantum_info.SparseObservable`, will (in the future) replace/supplement `SparsePauliOp` as the "abstract observable" input to `Estimator`.
-This will for construction, representation and transmission of one set of mathematical operations.
-It will be first be optimised for minimal memory usage.
+It will first be optimised for minimal memory usage.
 Its secondary concern will be to provide efficient mathematical manipulations.
 
 `SparseObservable` will _only_ store single-qubit terms that are not the identity (unlike `SparsePauliOp`).
@@ -367,9 +366,9 @@ For simplicity, I'm just talking about 64-bit systems here using 32-bit integers
 Let us assume that the number of non-identity single-qubit entries in any given term of an observable is, on average, a constant $k$ that depends on the `Estimator` problem.
 $k$ is non-strictly less than the number of qubits.
 
-`SparseObservable` uses 104 bytes on the stack, plus heap allocations of $\bigl((24 + 5k)\times\text{terms} + 8\bigr)$ bytes.  (The factor of $5k$ can be reduced to $4.5k$ with bit-packing of the alphabet terms.)
+`SparseObservable` uses 104 bytes inline, plus heap allocations of $\bigl((24 + 5k)\times\text{terms} + 8\bigr)$ bytes.  (The factor of $5k$ can be reduced to $4.5k$ with bit-packing of the alphabet terms.)
 
-A hypothetical Rust-space `SparsePauliOp` that uses maximal bit-packing of the symplectic arrays and merges the phases into the complex double-precision coefficients uses around 80 bytes of stack space, plus heap allocations of $\text{terms}\times(\text{qubits}/4 + 16)$ bytes.
+A hypothetical Rust-space `SparsePauliOp` that uses maximal bit-packing of the symplectic arrays and merges the phases into the complex double-precision coefficients uses around 80 bytes inline, plus heap allocations of $\text{terms}\times(\text{qubits}/4 + 16)$ bytes.
 
 The existing Python-space `SparsePauliOp` uses rather more constant-offset storage, plus heap allocations of $\text{terms}\times(2\times\text{qubits} + 17)$ bytes.
 
--- a/####-sparse-observable.md
+++ b/####-sparse-observable.md
@@ -9,9 +9,11 @@
 
 ## Summary
 
-Add a new Rust-backed type (`SparseObservable`) that will replace `SparsePauliOp` from a user perspective as the "natural" type of a single abstract observable to `Estimator`.
+Add a new Rust-backed observable `SparseObservable` to solve two scaling problems with `SparsePauliOp`.
 The new operator will have an extended alphabet of single-qubit terms, in order to avoid exponential explosion in representing projectors, and will be qubit-sparse in that it will not store explicit identities.
 
+This is expected to become part of the `Estimator` interface in the future, but the exact timing and plan for that is out of scope for this document.
+We still consider how `SparseObservable` could be used by `Estimator` as part of ensuring the interfaces are suitable.
 We separate the ideas of "representation of observables" from "representation of measurement bases", and use different objects to represent and manipulate these.
 The former of these is user-facing and will be part of interfaces (including across network boundaries), so needs to be API- and ideally ABI stable.
 The latter of these is an implementation detail of `Estimator`s or the circuit-knitting toolbox, so what is used can be changed later if needed for optimisation.
@@ -31,15 +33,19 @@ The latter of these is an implementation detail of `Estimator`s or the circuit-k
 - Primitives implementers will maintain a well-specified serialisable format for data transmission.
 - Primitives implementers will have access to high-performance classes for the operations they need.
 
+Note that these benefits can only be realised only once primitives _use_ this new type, which might require some `Estimator` interface evolution out-of-scope of this document.
+
 
 ## Design Proposal
 
-The new `SparseObservable` class is primarily designed to be the user-input format for observables in an `EstimatorV3` interface, but will also have mathematical operations defined for construction and manipulation.
-Since it will be used in the primitives interfaces, its data representation cannot change once in Qiskit.
-
+The new `SparseObservable` class is primarily designed to address scaling problems with `SparsePauliOp`.
+It will use an extended alphabet to avoid exponential memory explosions in `SparsePauliOp` when representing projection operators (like $`\lvert00\dotsm0\rangle\langle00\dotsm0\rvert`$), and will not store identities on qubits.
+Its storage format will be somewhat akin to a CSR/CSC matrix.
+Since it is intended to be used in stable interfaces, its internal data representation will be part of its public API, and cannot change once in Qiskit.
 
 In current Qiskit / primitives, the concept of "observable" (`SparsePauliOp`) is described in terms of the object that describes the necessary "measurement bases" (`PauliList`).
 With an extended observable alphabet, there can be a many-to-one mapping from observable terms to measurement bases ($Z$ and $\lvert 0\rangle\langle 0\rvert$ in observables both correspond to $Z$ in measurement), so there is a natural break between the two.
+The new operator will formalise this within its interfaces; it will have methods to produce the measurement bases needed to measure the terms.
 
 ### Definitions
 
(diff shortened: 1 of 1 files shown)


####-sparse-observable.md
```diff
@@ -0,0 +1,518 @@
...
+
+There are several use-cases for observables throughout the entire life cycle of "a user calls a primitive and gets a result":
+
+1. A user must be able to efficiently construct their operators, which may involve mathematical manipulations as they build up their problem.
+   This prioritises efficient construction and manipulation of the objects, while maintaining a clear mathematical meaning in the context of observables.
+
+2. All observables for an estimator pub / batch of pubs must exist simultaneously in memory on the client side and be transmitted over the wire, so must not use excessive memory.
+
+3. An estimator primitive (or circuit-knitting toolbox, etc) must be able to efficiently extract all the terms that correspond to unique measurements that must be made, across the entire broadcast `EstimatorPub`.
+   This is not necessarily the same set of mathematical terms as appear in a single operator, and the available groupings are dictated not just be individual mathematical operators or the `ObservablesArray`, but also in how it broadcasts against the `BindingsArray`.
```
blakejohnson, line 83:
```suggestion
   This is not necessarily the same set of mathematical terms as appear in a single operator, and the available groupings are dictated not just by individual mathematical operators or the `ObservablesArray`, but also in how it broadcasts against the `BindingsArray`.
```

####-sparse-observable.md
```diff
@@ -0,0 +1,518 @@
...
+
+
+### `SparseObservable`: representation of observables
+
+> [!NOTE]
+> **Summary**: A new observable class in `quantum_info` will represent Paulis and eigenstate-projectors, and store only non-identity terms.
+> This will be for users' benefit, and will be the "natural" input format of `Estimator`.
+
+A new class, `qiskit.quantum_info.SparseObservable`, will (in the future) replace/supplement `SparsePauliOp` as the "abstract observable" input to `Estimator`.
+This will for construction, representation and transmission of one set of mathematical operations.
```
blakejohnson, line 100: will be used for?

####-sparse-observable.md
```diff
@@ -0,0 +1,518 @@
...
+It will be first be optimised for minimal memory usage.
```
blakejohnson, line 101:
```suggestion
It will first be optimised for minimal memory usage.
```

####-sparse-observable.md
```diff
@@ -0,0 +1,518 @@
...
+
+* More flexibility in what can be represented in the operators.
+
+Cons:
+
+* Operations on the type will be more complex, and the general case will need to hold the Python GIL in Rust space to act on the observables.
+* More complex code is needed from Rust space to handle mathematical manipulations efficiently in the happy path of "no Python-space parametrisation".
+
+Jake: personally I'd avoid this unless we have a really strong compelling use-case for giving it first-class support.
+A user can always work around this simply by splitting the terms of their sum into different broadcast axes in the estimator pub, then calculating the sums themselves, which gives them far more freedom.
```
blakejohnson, line 477: I haven't seen strong use cases requiring parameterized observables, so I am fine leaving this out.

blakejohnson:
Looks like a strong proposal.

####-sparse-observable.md
jakelishman: I honestly have no idea what's going on here.  It feels like three completely disjoint thoughts - maybe I got distracted while rewriting it or something.  I deleted it - whatever I was trying to say should be covered by the rest of the section anyway.

####-sparse-observable.md
jakelishman: Done in 52eb1c1.

####-sparse-observable.md
jakelishman: Done in 52eb1c1.

####-sparse-observable.md
```diff
@@ -0,0 +1,517 @@
...
+The new operator will formalise this within its interfaces; it will have methods to produce the measurement bases needed to measure the terms.
+
+### Definitions
+
+There are several different things that could be "single" or an "atom" here, as there are many groupings.
+For clarity, we will use these definitions throughout:
+
+* **abstract observable**:
+  One mathematical operator that the user is requesting expectation values for.
+  This has the form $\sum_i c_i A_i\$ where $`\{c_i\}`$ are complex numbers and $`\{A_i\}`$ are some possibly over-complete spanning set of the operator space.
```
Cryoris, line 57:
```suggestion
  This has the form $\sum_i c_i A_i$ where $`\{c_i\}`$ are complex numbers and $`\{A_i\}`$ are some possibly over-complete spanning set of the operator space.
```
```suggestion
  This has the form $\sum_i c_i A_i\$ where $`\{c_i\}`$ are complex numbers and $`\{A_i\}`$ are some possibly over-complete spanning set of the operator space.
```

####-sparse-observable.md
```diff
@@ -0,0 +1,517 @@
...
+
+While the user input is a n-dimensional group of observable sums, the actual execution is serialised and only cares about measurement-basis transformation needed for any given multi-qubit term.
+See the "Detailed Design" section for more discussion on exactly how this grouping works.
+
+From Python-space, then, we produce an iterable of "preparation programs" (pairs of `(circuit, bindings)`), and for element of that we have an iterable of `SparseObservable` with we want to map to a group of groups of measurement bases.
+The iterable of preparation programs is produced in Python-space by n-dimensional slicing/broadcasting tricks.
+
+A core function of `Estimator` is to group terms that can be measured within the same execution.
+There are many ways to do this, and we do not want to tie the observable to one particular implementation.
+Qiskit will provide a function `SparseObservable.measurement_bases(*observables)` that takes an arbitrary number of `SparseObservable` instances and returns a set of the measurement bases needed to measure all terms.
```
Cryoris, line 155: This method would then just give all required measurement basis w/o any kind of optimization, which is left to other algorithms, right?

####-sparse-observable.md
```diff
@@ -0,0 +1,517 @@
...
+* the scalar element of `s_terms` is 4 bits, which requires 1 byte for alignment if not packed, or 0.5 bytes (average) if so.
+
+Consequently, the savings for bit-packing the `s_terms` vector is asymptotically _at most_ $`0.5 / 5 = 10\%`$, but in practice will be less than this.
+Bit-packing will also have some amount of performance impact, and would hinder the creation of cheap "term-view" objects within Rust space, since a term-based slice would not necessarily start on a byte-aligned `SingleBitTerm` values.
+
+The class will be able to present itself without data copying as a Python record type of a similar structure:
+
+```python
+class SparseObservableView:
+    num_bits: int
```
Cryoris, line 218: Maybe a too small comment for an RFC, but should this be `num_qubits` for consistency with the other observables & circuit?

####-sparse-observable.md
```diff
@@ -0,0 +1,517 @@
...
+* addition of two `SparseObservable`s
+* tensor product of two `SparseObservable`s
+* evolution of one `SparseObservable` by a single term
+* multiplication by complex scalars
+* structural equality of two `SparseObservable`s (structural not mathematical; it's highly inefficient to detect equality if the abstract terms form an overcomplete spanning set).
+
+There are several others that might be possible too.
+These operations could all fairly easily be supported:
+
+* Evolution of `SparseObservable` by another: this is completely doable, just naturally has quadratic complexity.
```
Cryoris, line 492: It comes up sometimes to compute the expectation value of the Hamiltonian squared in algos & applications (even in the estimator for the variance maybe?), so the would seem like an important feature

####-sparse-observable.md
jakelishman: Yeah, the purpose is to separate "grouping of measurement bases" (which primitives / users will likely want to configure) from "what are the measurement bases?".

####-sparse-observable.md
jakelishman: The names and whatnot can be bikeshed, but this discrepancy was purposeful in the RFC, at least: there's interest from the primitives in not having this tied to _qubits_ because of how they (eventually) plan to work dynamic circuits into the mixture.

That said, Ian and I talked about that fairly early on in the process, and it may well be something that's better done just by primitives-side documentation, and keep `num_qubits` as the term here, as you say, for consistency.

####-sparse-observable.md
jakelishman: Adding the expectation value computation for a given state is certainly easy enough, the trick is probably just finding some sensible representation of the statevector; we don't have the concept of a "sparse statevector" in Qiskit at the moment, and this operator is intended for numbers of qubits that dense statevectors can't represent.

####-sparse-observable.md
mrossinek: There are some use-cases around parameterized observables but I agree that I would not classify them as "strong", especially since there are alternative means by which to achieve the same end-result.

That said, I wonder if it will be possible to have some Python-level utilities (or maybe alternative classes) that could make working with parameterized observables a little bit easier. I am not sure what exactly that would look like, but am wondering if @jakelishman could comment on the feasibility of something like this.

####-sparse-observable.md
```diff
@@ -0,0 +1,517 @@
...
+* Tidy-up structural compaction of the operator: summing all terms that share the same abstract operator, removing zeros at some specified tolerance.
```
mrossinek, line 493: I think this would be somewhat crucial, especially if we consider construction of operators to happen in an "iterative" manner by means of a sequence of mathematical operations.
The existing `SparsePauliOp` already has a part of its existing `.simplify` routine implemented in Rust, but performance still leaves room for improvement. I wonder how much this new operator completely implemented in Rust would benefit purely from the language advantage? It might be difficult to predict though :thinking: 

At the same time, could the CSR-like structure allow an alternative (read: more performant) approach for an implementation of `.simplify`?

Just thinking out loud here...

mrossinek (commented):
I like the proposal a lot :+1: 

Just left two comments on aspects that stood out to me as noteworthy.

####-sparse-observable.md
jakelishman: For background, the high-level algorithm of `SparsePauliOp.simplify` is basically:
1. start with an empty hashmap of `pauli: coeff`
2. for each term of the sum, add the term into the hashmap, summing the coeffs if there's a match
3. create a new `SparsePauliOp` with each term in the hashmap, if the coeff is sufficiently far from 0

The _way_ we do that has a fair amount of fiddling around the edges that's making it less efficient than it could be, but asymptotically, the runtime complexity is already the best I can think of.

That said, the complexity still scales as $`\mathcal O(\text{qubits}\times\text{terms})`$.  The CSR-like form would be effectively the same, and so have the same asymptotic complexity. As a rule of thumb, though, it would be faster when there's less memory used (i.e. individually sparse terms), simply because there's no real mathematical structure / algorithmic trickery going on here, and the limit is mostly the iteration speed over all stored data.

For operator construction, one thing (mostly unrelated to this comment) to highlight: the way I've written `SparseObservable` here, it's growable in place, which means for some operations (`+` being a notable one), we can do it in-place and growing, which is very nice for iteration.  `evolve` is harder to do in-place, though.

####-sparse-observable.md
ihincks: Calling them qubits is fine with me. The idea with dynamic circuits would be to attach observables to slices of measurement registers rather than to the entire circuit, with measurement terms prescribing instructions to insert before those registers. Therefore multiple qubits in an `SparseObservable` could in principle correspond to a single qubit, at the quantum programmer's discretion.

Since we are not there yet, and even when we get there it would be a point of minor concern, I'm not fussed.

####-sparse-observable.md
Cryoris: Hmm but if we square a Hamiltonian we might have to measure in new bases, e.g. if `H = X -> H^2 = I` we'd have to measure in Z-basis for the expval of `H^2` (ok maybe a bit basic this example... 😄). Or maybe I didn't correctly understand your proposal 🤔

jakelishman: Oh sorry, I missed the word "squared" in your answer.

If we have some representation of a state, I think the expectation value of the square of a Hamiltonian might shake out neater in an API if we cast that problem to "evolve the (sparse) state by the operator, then take the inner product with itself"?

pedrorrivero (commented):
I really like this (much needed) RFC, awesome work @jakelishman @ihincks !

####-sparse-observable.md
```diff
@@ -0,0 +1,517 @@
...
+The iterable of preparation programs is produced in Python-space by n-dimensional slicing/broadcasting tricks.
+
+A core function of `Estimator` is to group terms that can be measured within the same execution.
+There are many ways to do this, and we do not want to tie the observable to one particular implementation.
+Qiskit will provide a function `SparseObservable.measurement_bases(*observables)` that takes an arbitrary number of `SparseObservable` instances and returns a set of the measurement bases needed to measure all terms.
+A measurement basis is a Pauli string.
+
+Since the number of measurement bases will be (non-strictly) smaller than the total number of terms across all observables, and because only 2 bits of information per qubit is necessary to define the basis, there is not expected to be immediate memory concerns with a representation of this.
+Qiskit already has `PauliList` that can serve this purpose; it _could_ be bit-packed to use 8x less memory, but this can be done as a follow-up optimisation if it becomes a bottleneck.
+From this point, we can continue to use `PauliList.group_qubitwise_commuting`, or any other future grouping function.
```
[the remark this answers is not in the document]
jakelishman: The only part of `SparsePauliOp` that causes serious immediate memory problems is that some things that are efficient to measure are not efficient to represent.  For example, the all-zeros projector state takes linear complexity/space to measure (it's an all-Z basis) but needs $2^n$ terms in `SparsePauliOp` to represent.  Since, for the measurement, you just need to know which basis to take your counts in, the information about "what to measure" (including mitigation) can be stored efficiently by `PauliList`, and you reconstruct the requested observables later from the original `SparseObservable`.

Bit-packing `PauliList` can reduce its memory usage by a factor of about 8, but that's just a scaling factor.  We already know we _must_ be able to do operations that are linear in the number of qubits (or how would we twirl?), and `PauliList` can represent all the measurement bases we realistically care about for error-mitigation purposes for the next while.

All that said, how a primitives implementation actually manages its error mitigation is entirely up to it, and this isn't fixed by any public interface.  So any primitive can use whatever they like in the backend to do this task; the point about `PauliList` is mostly just showing that we don't immediately need any new object.

####-sparse-observable.md
```diff
@@ -0,0 +1,517 @@
...
+The particulars of `Parameter` and `Value` and the containing mapping are irrelevant to this RFC.
+
+`SparseObservable` will support some set of mathematical operations.
+At a minimum, the following will be supported:
+
+* addition of two `SparseObservable`s
+* tensor product of two `SparseObservable`s
+* evolution of one `SparseObservable` by a Pauli string ($A' = P A P^\dagger$ for `SparseObservable` $A$ and Pauli $P$)
+* multiplication by complex scalars
+* structural equality of two `SparseObservable`s (structural not mathematical; it's highly inefficient to detect equality if the abstract terms form an over-complete spanning set).
```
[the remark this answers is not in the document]
jakelishman: Sure, `apply_layout` is easy enough to do.  Longer term, the story around compilation of observables will probably change a bit in Qiskit to be more streamlined, but for now we'll keep consistency with `SparsePauliOp` on that.

Evolution is mathematically sound for many operators, the tricks are mostly around finding nice representations of more complex objects for the API surface.

####-sparse-observable.md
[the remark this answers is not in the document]
pedrorrivero: Thanks @jakelishman! I was thinking more on all the design decisions to avoid explicit identities, but I see your point that this is not the biggest concern 🙂

ihincks (approved):

####-sparse-observable.md
jakelishman: We can call this `qubits` in the actual implementation, no trouble.

blakejohnson (approved):

####-sparse-observable.md
Cryoris: Hm I'm not sure you can do that since you'd actually have to measure in different bases.. but we can also discuss this later 🙂

pedrorrivero (approved):

mrossinek (approved):

Cryoris (approved):

PR #77, opened 2025-04-04
`BackendV3`

ElePT:
This RFC summarizes a proposal for `BackendV3` that has been discussed for a few months already. The proposal is simple, but the discussions around the topic are not. The document focuses on adding historical context and capturing arguments for and against to clarify the motivation for the proposal.


0023-backend-v3.md
```diff
@@ -0,0 +1,239 @@
...
+The `Target` class is a container for the different hardware specifications that model a QPU. It was modelled following the requirements of the Qiskit transpiler. It's designed to be vendor-agnostic, but allows for vendor-specific extensions.
+
+The `Target` defines 3 categories of information in its attributes:
+
+1. **Required:** necessary to define a fundamental instance. Qiskit (and in particular, the transpiler) relies on their definition. These are:
+    - number of qubits
+    - supported instructions, as tuples of `(instruction, qargs)`
+
+2. **Optional:** this information is not necessary, but if provided, Qiskit will offer tools to leverage this information and reason about it. These are:
+    - `QubitProperties`
```
mtreinish, line 155: One aspect that we got wrong in `BackendV2` is that `QubitProperties` started as a property of the `Backend` and not the `Target`. At the time the target was only thought of as what was needed for transpilation and we weren't consuming qubit properties in the transpiler. So we ended up in a weird state where it's optional in the backend and optional in the target to support people implementing `BackendV2` with qubit properties prior to qiskit-terra 0.20.0. I think in BackendV3 we should fix this and make it optional on the `Target` only and the `BackendV3` side alias a target query only.

0023-backend-v3.md
```diff
@@ -0,0 +1,239 @@
...
+| **RFC #**         | ####                                         |
+| **Authors**       | Elena Peña Tapia (ept@zurich.ibm.com), Matthew Treinish (matthew.treinish@ibm.com)   |
+| **Deprecates**    | RFC that this RFC deprecates                 |
+| **Submitted**     | YYYY-MM-DD                                   |
+| **Updated**       | YYYY-MM-DD                                   |
+
+## Summary
+Backends are the "de facto" abstractions for quantum devices in the Qiskit ecosystem, extending beyond a simple combination of their components. The current Qiskit backend model, however, clashes with the execution model of primitives. This is a proposal to add a new `BackendV3` class that fills the gap in the current Qiskit backend + execution model.
+
+> Instead of a required `backend.run()`, the proposal for `BackendV3` implementations is to offer **primitive factory methods**: `backend.sampler()` and `backend.estimator()` that return their respective instantiated primitive classes (details TBD), as well as an optional `backend.run()` for 3rd party and local simulation backends (for backwards compatibility and testing, up to discussion).
```
mtreinish, line 14: I think one big detail to discuss/decide on would be on how to handle primitive versioning. As the primitives interface is independently versioned and typically a provider only implements a single version per release is there a mechanism we want to define here around dealing with that.

The approaches I can think of are either we are prescriptive in the versioning, so say explicitly `BackendV3.sampler() -> SamplerV2` and then if/when there is a SamplerV3 we have to make a `BackendV4` to bump the return type to use `SamplerV3`. The other is to leave it opened ended and say that the `BackendV3.sampler() -> Sampler` where it is any version and leave it up to the user to handle the version management. The third would be version the methods, which I think I like the least, e.g. `backend.sampler_v2()`.

0023-backend-v3.md
```diff
@@ -0,0 +1,239 @@
...
+## Motivation
+
+### Why are we doing this? (With a bit of History)
+
+The Qiskit `Backend` object was originally introduced to provide an abstract model of a **physical quantum computer**, enabling software to interact with it. There are two key types of interaction with a `Backend`:
+
+1. **Transpilation** -> "we need to access hardware constraints to optimize circuits (to run on a QPU)"
+2. **Execution** -> "we need an interface for sending circuits to the QPU and receiving an output"
+
+The original `Backend` interface (`BackendV1`) was designed primarily for circuit execution, the most fundamental ("basic") interaction with quantum computers. However, the need for improved transpilation support led to the introduction of `BackendV2`, and incorporating an abstraction for the "QPU", the `Target`.
```
mtreinish, line 44: The original backend interface was actually `BaseBackend` (i.e. `BackendV0`), which was basically `BackendV1` but with 2 key differences, `BaseBackend` took a `qobj` as an input instead of a `QuantumCircuit` on `.run()` and it was also explicitly tied to the IQX API formats. For `BackendV1` I made it explicitly decoupled from IBM's REST API payload format but still using the same format at the time, this was to decouple the Python code from the API format to allow changes on either side, and removed the use of qobj from the `run()`.

0023-backend-v3.md
```diff
@@ -0,0 +1,239 @@
...
+1. Providing a interface general enough to support any vendor (**vendor-agnostic**)
+2. Setting a level of abstraction that allows the **physical hardware to evolve** without affecting the abstract object
+
+Both these goals have been fulfilled, with a growing list of 3rd-party providers that consume the `Backend` interface, which has seamlessly supported hardware changes in basis gates, couplers, noise definitions, introduction of qubit resets, and more. The goal for `BackendV3` is to maintain these design principles.
+
+### What will be the outcome?
+
+A new `BackendV3` class. This class should be almost 100% backwards-compatible with the current transpilation stack (for which it must contain a `Target`), but not require a `.run` method, and instead provide primitive-friendly alternatives.
+
+Instead of a required `backend.run()`, the proposal for `BackendV3` implementations is to offer **primitive factory methods**: `backend.sampler()` and `backend.estimator()` that return their respective instantiated primitive classes (details TBD), as well as an optional `backend.run()` for 3rd party and local simulation backends (up to discussion).
```
mtreinish, line 65: I think this is very important aspect, the difference between the migration from `BaseBackend` -> `BackendV1` and `BackendV1` -> `BackendV2` is that moving to `BackendV1` was a simple migration it changed very little. If there was only one lesson from `BackendV2` is that changing too much at once makes it exponentially harder to get everyone to migrate. So keeping this small and not breaking existing implementations so the migration is trivial this is a better way to evolve the interface.

0023-backend-v3.md
```diff
@@ -0,0 +1,239 @@
...
+- Using examples to introduce new features
+- Implementation and Migration path with associated concerns
+- Communication of features and changes to users
+
+Focus on giving an overview of impact of the proposed changes to the target
+audience. -->
+
+### BackendV2 - the existing backend model
+
+Internal communication guidelines establish the following definitions:
```
mtreinish, line 94:
```suggestion
IBM Quantum uses the following definitions:
```

0023-backend-v3.md
```diff
@@ -0,0 +1,239 @@
...
+
+This sums up the change from having one mandatory implementation of the execution step (`.run()`) to having three optional implementations: 
+
+- `.sampler()`
+- `.estimator()`
+- `.run()`
+
+> A point that should be clarified is what we'd like the interface to require. With requiring `.run()` out of the question, should we make them all optional and not require backends to be executable? Should we require at least a `.sampler()` or a `.estimator()` method?
+
+It seems like not requiring an execution step is the most flexible alternative, as the need for a container that includes the target and the extra functionality still justify the need for having a backend object in these cases.
```
mtreinish, line 129: I think making all 3 optional is the best choice here. So you can have an "execution-less" backend object that only exposes a model for compilation. We also should be explicit about the role of `run()` as this often gets muddled because people think it's just counts. `run()` is designed to be very free form where you get a `QuantumCircuit` in and a `Result` out. `Result` is purposely unrestricted, there are **no** requirements on the data it contains and the implementation details are left up to the implementer/vendor. There is a bias/convention towards returning `Counts` per circuit input in the job as that is what IBM backends did at the time the `Result` object was created and there are some formatting expectations if using those count mechanisms. But you don't have to use them at all. Aer is a good example of something that uses this feature of `Result` as it attaches other circuit output in the result object. For example, the various custom `Save*` instructions it offers (like `SaveExpectationValue` or `SaveStatevector`) which attach extra data fields for the result of these save instructions.

This differs from the primitives which are very prescriptive in how the execution runs and the result formats. I think they all have their place depending on the implementation, but we want to make sure either here or in the `BackendV3` docs we make it clear the role each method plays.

mtreinish (commented):
Thanks for doing this @ElePT this is a great start. I did a quick read through it and left some initial thoughts inline.

0023-backend-v3.md
ElePT: I had the same thought, I think I added somewhere that I think that the method should be independent of the primitives version, so I was thinking of `BackendV3.sampler() -> Sampler`. I hadn't considered the idea of versioning the methods, and I think that could work too. There are pros and cons to each alternative, I think I will make a table with them.

nonhermitian:
What is the benefit of having a `Sampler` and `Estimator` class here rather than just cutting them out entirely and just doing `backend.sample(qc)` and `backend.estimate(qc)`?  If you go through the online Primitives examples the only thing they ever do is immediately call `*.run(qc, parms)` on the `Sampler` and `Estimator`; they are just middle man objects with no intrinsic value for the user.

0023-backend-v3.md
ElePT: Added in [0427d3d](https://github.com/Qiskit/RFCs/pull/77/commits/0427d3da30df0bf05fcf405b92120e4cbfa4edc5)

0023-backend-v3.md
ElePT: Added in [0427d3d](https://github.com/Qiskit/RFCs/pull/77/commits/0427d3da30df0bf05fcf405b92120e4cbfa4edc5)

0023-backend-v3.md
ElePT: Added in [0427d3d](https://github.com/Qiskit/RFCs/pull/77/commits/0427d3da30df0bf05fcf405b92120e4cbfa4edc5)

ElePT:
> What is the benefit of having a `Sampler` and `Estimator` class here rather than just cutting them out entirely and just doing `backend.sample(qc)` and `backend.estimate(qc)`? If you go through the online Primitives examples the only thing they ever do is immediately call `*.run(qc, parms)` on the `Sampler` and `Estimator`; they are just middle man objects with no intrinsic value for the user.

From a developer and API perspective these middle man objects have a clear purpose and intrinsic value. They maintain the separation of concerns and deal with different server-side APIs. Primitives are not lightweight, and having a single backend-sampler-estimator would be chaotic and difficult to manage. The decoupling allows for better maintainability, stability, and less breaking changes overall. They also already exist and are well established and documented, and previous migrations have shown that drastic changes such as "wiping out" objects are increasingly difficult to carry out. 

From a user perspective, the primitive examples also instantiate the primitives and define and set execution mode and options (shots, error mitigation and the works). If we were to merge backend and both primitives into a single backend-sampler-estimator, let's imagine that we are using `qiskit-ibm-runtime`, call `service.backend("ibm_fez")`, and we want to run an estimator in batch mode with 10k shots and a rep delay of 10^-4. What would be the user benefit of doing something like:

```
from qiskit_ibm_runtime import QiskitRuntimeService

service = QiskitRuntimeService()
backend = service.backend("ibm_fez")
backend.update_estimator_mode("batch")
backend.update_estimator_options({"shots":1024, "rep_delay": 1e-4})
job = backend.estimate([pub])
```

instead of the proposed

```
from qiskit_ibm_runtime import QiskitRuntimeService

service = QiskitRuntimeService()
backend = service.backend("ibm_fez")
estimator = backend.estimator(mode="batch", options={"shots":1024, "rep_delay": 1e-4})
job = estimator.run([pub])
```
?

ElePT:
> So the primary benefit is that I do not have to care/learn about another object. Especially one that is primarily an IBM Quantum construct. i.e. no one else has such things

My argument is not to introduce something new, but rather not disrupt what already exists. We are not building a new model from scratch, primitives already exist in our ecosystem. We are not picking between two equal options and my stance here is that removing them at this point would cause more harm than benefit. This is why, instead of removing the notion of primitives, this is a proposal to streamline their creation and smoothen the curve for users. With `BackendV3` users don't need to pick a primitive, it's just something that is returned from the backend and users can run on demand. All without disrupting existing workflows.

nonhermitian:
At the end of the day do what you will I guess.  By abstracting away the execution mode, you are going in the right direction, but somehow are not willing to go the final step to backend nirvana.  Just seeing the notion of `estimator_options` is disappointing as it indicates that people have not learned from previous experience with 14 different options classes, and think that runtime execution modes are independent constructs.  As it stands now, there is little of value in this V3 proposal for end users to be happy about.  Rather it seems to just be porting poor choices in the runtime interface over to Qiskit.  I am happy though that at least `backend.run` is still there so those of us that try to write code that is provider agnostic still have a common point to work with (provided one just calls the sampler within the `backend.run` call)

ElePT:
We are working with examples that are too overfitted to the `IBMBackend` when discussing something abstract and general. The `BackendV3` interface doesn't limit how a downstream implementation can deal with things like "options", this is up to the specific implementation. Things like whether "rep_delay" is set at a backend level or primitive level (whether it's called `estimator_option` or `backend_option` or `execution_option`) is up to the implementer of the interface, all of them are possible in `BackendV3`. The only thing that the interface dictates is the `sampler()` factory instead of the `sample()` method. The base primitives already exist in Qiskit so we aren't porting anything from runtime. We are just unifying what exists in Qiskit.

0023-backend-v3.md
```diff
@@ -0,0 +1,269 @@
...
+estimator = backend.estimator(mode="batch", options={"shots":1024})
+```
+
+instead of:
+```python
+from qiskit_ibm_runtime import QiskitRuntimeService, Estimator
+
+service = QiskitRuntimeService()
+backend = service.backend("ibm_fez")
+estimator = Estimator(backend=backend, mode="batch", options={"shots":1024})
```
ihincks, line 32: qiskit-ibm-runtime estimator currently has this construction signature:
```python
    def __init__(
        self,
        mode: Optional[Union[BackendV2, Session, Batch, str]] = None,
        options: Optional[Union[Dict, EstimatorOptions]] = None,
    ):
```

0023-backend-v3.md
```diff
@@ -0,0 +1,269 @@
...
+
+Making all 3 methods optional seem like the best choice because of its flexibility: it enables all three execution options as well as an "execution-less" backend object that only exposes a model for compilation (transpilation). This can be different from just using a `Target` because of the additional functionality backend implementations provide.
+
+2. **The role of .run, why is it valuable** 
+
+The role of `run()`, as defined in `BackendV2` often gets immediately translated to "return counts", which is more restrictive than what the interface dictates. `run()` is designed to be free form, the only requirements are that the input is  a `QuantumCircuit` and the output is a `Result`. `Result` is purposely unrestricted, there are no requirements on the data it contains and the implementation details are left up to the implementer/vendor. There is a bias/convention towards returning `Counts` per circuit input in the job for historical reasons, and there are some formatting expectations if using those count mechanisms. But the output of `.run()` doesn't have to be "counts".
```
ihincks, line 141: I have trouble finding the value of abstracting an execution mode whose implementation is optional, and when implemented, whose output is free-form. It seems like the real motivation for keeping .run in the abstraction is for historical continuity.

0023-backend-v3.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### 3. Extra functionality (small changes in BackendV3)
+
+As mentioned initially, this category includes both optional and custom functionality that is currently present in `BackendV2` implementations. The difference between optional and custom is that Qiskit offers tools to reason about optional functionality, while custom functionality requires custom tools to handle it (similar philosophy to the `Target` class).
+
+**Optional Functionality** -> `QubitProperties`
+
+For historical reasons, `QubitProperties` remains an optional property on both the `Target` and `BackendV2` objects. This adds redundancy and ambiguity, as any functionality that uses `QubitProperties` must query both and establish rules on what takes precedence over what. This can be easily fixed in `BackendV3` by no longer storing `QubitProperties` in the backend, and keeping them exclusively as an optional `Target` property. Querying these properties would still be possible in `BackendV3` for backwards compatibility by aliasing the target query.
```
ihincks, line 213: Aliasing them would solve the problem of them being out of sync, but it doesn't seem to address the ambiguity problem, which I view as the real problem, as I suspect that the average user will not understand that it's an alias, and still wonder why the heck the same information seems to be in two places.

I would favor making a breaking change here.

0023-backend-v3.md
```diff
@@ -0,0 +1,269 @@
...
+For example, `IBMBackend` exposes the following hook to handle fractional gate support, that overwrites the default translation plugin with either `"ibm_dynamic_circuits"` or `"ibm_fractional"`.
+
+```python
+class IBMBackend(...):
+  ...
+  def get_translation_stage_plugin(self) -> str:
+    """Return the default translation stage plugin name for IBM backends."""
+    if not self.options.use_fractional_gates:
+      return "ibm_dynamic_circuits"
+        return "ibm_fractional"
```
ihincks, line 230: missing an `else` or something?

jyu00:
From Qiskit Runtime's perspective, our users just went through great pain to migrate from `backend.run` to the primitives. Personally I don't see enough value out of this proposal to force them to change their code yet again to go from `Estimator(backend)` to `backend.estimator()`.

ElePT:
> From Qiskit Runtime's perspective, our users just went through great pain to migrate from `backend.run` to the primitives. Personally I don't see enough value out of this proposal to force them to change their code yet again to go from `Estimator(backend)` to `backend.estimator()`.

`Estimator(backend)` and `backend.estimator()` are complementary, they can both be supported without forcing users to change their code. This is part of the backwards compatibility of the proposal. None of the proposed changes are breaking changes.

jyu00:
> `Estimator(backend)` and `backend.estimator()` are complementary, they can both be supported without forcing users to change their code. This is part of the backwards compatibility of the proposal. None of the proposed changes are breaking changes.

If we don't plan to deprecate `Estimator(backend)`, then it's hard to justify putting in the work to implement this new interface. Especially if we want to change all docs/tutorials etc to push the new interface. Having multiple ways of doing the same thing often raises questions and confusion.


> Choosing the right primitive (especially when switching between local simulation and HW execution) is one of the biggest pain points for users and the biggest user-facing motivation for this proposal.

Qiskit Runtime has a [local mode](https://docs.quantum.ibm.com/guides/local-testing-mode) that's meant to allow users to switch between local simulation and IBM HW without changing the code besides the `backend`. It's not perfect, but wouldn't it be better to put the effort into improving it instead?

PR #78, opened 2025-08-14
Introduce new Executor primitive incorporating samplomatic semantics

delapuente:
This RFC proposes the introduction of a new Executor primitive into Qiskit as a vendor-agnostic substrate for execution. The `Executor` provides three core components—`Executor`, `ExecutorProgram`, and `ExecutorResult`—that capture the relationship between a programming model engine, its semantics, and its results.

The immediate motivation is to make randomization semantics for error mitigation explicit and first-class in Qiskit. Today, such workflows are implicit and opaque; this RFC standardizes how users express their intent through annotations and portable representations. This enables reproducibility, fine-grained control, and composability, while letting providers implement and optimize their own expansion engines.

This RFC does not introduce a specific engine such as IBM’s Samplomatic into Qiskit. Instead, it defines the substrate and user-intent representation, leaving execution engines to providers. Follow-up RFCs may propose additional semantics, a reference `Executor`, or reimplementations of existing primitives (e.g., `Sampler` and `Estimator`) atop this foundation.


0023-low-level-sampler.md
```diff
@@ -0,0 +1,203 @@
...
+backend = service.backend("test_heron")
+
+# Build a circuit for which we need to learn the noise.
+circuit0 = QuantumCircuit(2)
+
+with circuit0.box([Twirl(), InjectNoise("my_noise")]):
+  circuit0.cx(0, 1)
+
+with circuit.box([Twirl(), BasisTransform(ref="my_basis")]):
+  circuit.measure_all()
```
garrison, line 110: I believe you must have meant

```suggestion
with circuit0.box([Twirl(), BasisTransform(ref="my_basis")]):
  circuit0.measure_all()
```

0023-low-level-sampler.md
```diff
@@ -0,0 +1,203 @@
...
+  basis_transforms={"my_basis": "XX"},
+  cast_result_into_noise_map=True
+  cache_as="my_noise"
+)
+
+sampling_program = QuantumProgram(shots=1024)
+sampling_program.append(
+  circuit0,
+
+)
```
garrison, line 133: It looks like some arguments are missing to this call to `.append()`.

garrison (commented):
:+1:

0023-low-level-sampler.md
```diff
@@ -0,0 +1,188 @@
...
+| **Submitted**     | 2025-07-30                                   |
+| **Updated**       | YYYY-MM-DD                                   |
+
+## Summary
+In Qiskit, the quantum computer interface is built around primitives, which are
+the fundamental operations executable on a quantum circuit. Currently, Qiskit
+offers two such primitives: the Sampler and the Estimator. The Estimator
+leverages advanced error mitigation techniques to enhance the quality of
+results from noisy quantum computers. On a noiseless quantum computer, the
+Estimator could instead be constructed atop the Sampler, simplifying its role.
```
blakejohnson, line 17: The purpose of the Estimator extends beyond just mitigating noise. It also adds a layer abstraction around the concept of estimating observables, allowing implementors of Estimator interfaces to offer different methods for observable estimation (e.g. classical shadows) while preserving a common interface.

0023-low-level-sampler.md
```diff
@@ -0,0 +1,188 @@
...
+will gain a clearer and more composable model for error mitigation workflows,
+making it easier to experiment, reproduce results, and share techniques.
+
+## Design Proposal
+
+The intent of the following listing is to demonstrate how a user could leverage
+the new low-level interface to implement a client-side basic estimator.
+
+```python
+from qiskit_ibm_runtime import QiskitRuntimeService, Session
```
blakejohnson, line 76:
```suggestion
from qiskit_ibm_runtime import QiskitRuntimeService, Session, NoiseLearner
```

blakejohnson (changes requested):
This RFC seems to introduce 3 concepts:
1. Sampling graphs (`samplex`)
2. `Executor` primitive
3. `QuantumProgram` data structure

However, (2) and (3) aren't described at all beyond their appearance in the example. (1) gets the most motivating text, but the RFC doesn't actually say enough about what it is beyond something that produces circuit variations.

In short, this looks rather incomplete at this stage.

0023-executor-primitive.md
```diff
@@ -0,0 +1,188 @@
...
+## Summary
+In Qiskit, the quantum computer interface is built around primitives, which are
+the fundamental operations carried by a quantum computer on a quantum circuit.
+Currently, Qiskit offers two such primitives: the Sampler and the Estimator.
+The Estimator leverages advanced error mitigation techniques to enhance the quality
+of results from noisy quantum computers. On a noiseless quantum computer, the
+Estimator could instead be constructed atop the Sampler, simplifying its role.
+However, given current noise levels, robust error mitigation is essential,
+requiring executing thousands or even millions of circuit variations to implement
+certain mitigation protocols. Transmitting these variations across cloud environments
```
ihincks, line 20: Here we should distinguish between the estimator abstraction, which says nothing about mitigation, and our IBM hardware implementation---the primitives service---that has options to enable it.

0023-executor-primitive.md
```diff
@@ -0,0 +1,188 @@
...
+
+# This proposal assumes the semantics implemented in samplomatic will become
+# part of qiskit semantics at some moment.
+from qiskit import samplex
+from qiskit.samplex.annotations import InjectNoise, Twirl
+
+# The proposal wants to acknowledge the transtionary reality in which the
+# primitives continue being the entry point for operating the quantum computer
+# and this is a low level interface for advanced usage.
+from qiskit.primitives.implementation import QuantumProgram, Executor
```
ihincks, line 89: I expect that the `Executor` will need to come from `qiskit-ibm-runtime` instead: it needs to do things like POST to our service.

Instead, base class like BaseExecutor could live here.

I'm also wondering how this might all fit in with BackendV3, @ElePT , @jakelishman , @mtreinish : we might consider not having an `Executor` at all and reverting to something more like `backend.run()`. In this case we might need to have different semantics for sessions/batches than we have today.

0023-executor-primitive.md
```diff
@@ -0,0 +1,188 @@
...
+variations locally, reducing network load while preserving user control over
+error mitigation techniques.
+
+## Questions
+Open questions for discussion and an opening for feedback.
+TBD
+
+## Future Extensions
+
+The introduction of a low‑level sampler is only the first step. The long‑term
```
ihincks, line 171:
```suggestion
The introduction of a low‑level sampler (called an executor) is only the first step. The long‑term
```

0023-executor-primitive.md
ElePT: > I expect that the Executor will need to come from qiskit-ibm-runtime instead: it needs to do things like POST to our service.

They need to, same as sampler and estimator.

> I'm also wondering how this might all fit in with BackendV3, @ElePT , @jakelishman , @mtreinish : we might consider not having an Executor at all and reverting to something more like backend.run(). In this case we might need to have different semantics for sessions/batches than we have today.

I think this requires understanding how this proposal affects the "primitives model" of execution we currently have. Will this replace the primitives? complement the primitives? I don't know how we could reasonably justify reverting to `backend.run()` if the `sampler` and `estimator` primitives continue existing...

delapuente: Ok, I'm going to try to imitate the RFC for the [Estimator](https://github.com/Qiskit/RFCs/blob/master/0015-estimator-interface.md) in which the imports are not displayed, and elaborate on the concepts, then we can think of a strategy for adding this to Qiskit.

0023-low-level-sampler.md
```diff
@@ -0,0 +1,188 @@
...
+with circuit.box([Twirl(), BasisTransform(ref="my_basis")]):
+  circuit.measure_all()
+
+# Extract layers of the circuit
+layers = find_unique_layers(circuit)
+
+# Prepare circuit for executing
+template, samplex = samplex.build()
+
+with Session(backend=backend) as session:
```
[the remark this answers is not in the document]
delapuente: This is now addressed in its own section.

0023-executor-primitive.md
```diff
@@ -0,0 +1,335 @@
...
+  # Instantiate the programming model primitive
+  executor = Executor(backend)
+  
+  # Learning noise
+  noise_learner = NoiseLearningProgram(circuit_layers)
+	job = executor.run(noise_learner)
+  result = job.result()
+  
+  # Extrac noise map
+  noise_map = { layer: model for zip(circuit_layers, result }
```
SamFerracin:
```suggestion
  # Extract noise map
  noise_map = { layer: model for zip(circuit_layers, result) }
```

0023-executor-primitive.md
```diff
@@ -0,0 +1,335 @@
...
+important for implementers to address in provider-specific documentation.
+
+## Alternative Approaches
+An alternative to introducing the **samplex** DAG is to generate all circuit
+variations entirely on the client side and transmit them to the backend for
+execution. This approach offers maximum flexibility and complete control over
+how variations are produced, since the backend would simply execute the
+provided circuits without influencing their structure. Users could implement
+any custom error mitigation workflow they wish, with no restrictions imposed by
+backend capabilities or vendor‑specific optimizations.
```
SamFerracin: There is be another, more efficient alternative to "[generating] all circuit
variations entirely on the client side," which is to simply generate templates and parameters. However, a benefit of the `samplex` is that it allows generating parameters "on the fly:" vendors can set up their stack in such a way that their QPU can execute a set of parameters while a new set of parameters is being generated by the samplex. This can significantly speed up execution, with respect to both alternatives

delapuente: Did not realize passing the samplex enabled lazy generation. Thank you.

